PR #4, opened 2023-03-13
docker-compose and docker images working

RaceFPV:
adds support for docker-compose, docker builds, etc
docker-compose will create the mongodb container automagically and set it up
the docker version is a unified app (backend and frontend merged into one server)


Sinetheta:
Thanks, I can't get this to run on my M2 mac just yet, but I think if we get CI working and some tests running it will help us figure out which node versions to target then I can worry about Mac world.

PR #25, opened 2023-04-09
Update URL on website actions

Sinetheta:
This uses `location.replaceState` to update the URL during interactions which result in major app state change.

1. NEW blueprint clears the url
2. OPEN blurprint from browse updates the url
3. SAVE blueprint updates link to that blueprint
4. UPLOAD clears the url

This makes the "get shareable link" mostly redundant, but until we replace it with a more robust "share" system (eg: copy to clipboard, send to popular apps) I don't think it's hurting anyone to leave it in.


Sinetheta:
Unfortunately the places I had hooked in (closer to the cause than the effect) led to undesirable side-effects. So I've moved all these url updates to replaceState in the `BlueprintService`

Issue #1, opened 2023-03-09
B key for build please


Wizeowel:
Original project seemed to have a fix for this but it never went live.
https://github.com/simonlourson/blueprintnotincluded/pull/8

RaceFPV:
This has been confirmed working in the latest release today based on PR#13

Wizeowel:
Good work, thanks however unfortunate side effect is I can't type the word 'Incubator' into an Info tile. Please suppress this functionality while in a text box.

RaceFPV:
reopening after verifying unintended behavior bug

Ozzymud:
Adding onto unintended issue, pressing "b" while editing an info tag ends the edit.

Sinetheta:
Thanks for fixing this Race!

Looks like everything is working now, so I'll close the issue. Please re-open if anything comes up.

PR #38, opened 2024-03-09
Fix unintended tool switch when typing

SunnyHuangCodebase:
Resolves issue #1

Bug: Pressing the "b" key erroneously activates the build tool when user is attempting to type into a textbox.

Fix: Ignore "b" key presses when typing into `<textarea>` and `<input>` elements.


Sinetheta:
Thanks for the submission @SunnyHuangCodebase sorry I didn't notice it earlier. The "Delete" path also suffers the same problem but I'll merge this now and look into the logic where the handler is applied.

Sinetheta:
I think CI didn't run because of a setting on this repository. The docs make it sound like I should be able to "Approve and run" but that option doesn't appear. Maybe because I took so long to respond that it's gone?

https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

Issue #34, opened 2023-05-05
Automation wires appear under automation items


Ghent:
When automation items are present (gates, etc) the wires are displayed underneath, even in the automation overlay.  This is makes it very difficult to view how automation is connected.

![Screenshot 2023-05-05 144556](https://user-images.githubusercontent.com/47145/236543306-7d048d86-f59b-4680-b379-b527e6490ff8.png)
![Screenshot 2023-05-05 144655](https://user-images.githubusercontent.com/47145/236543311-c1d50e98-2ac2-4ae4-b10b-2152660700f9.png)

andrewvaughan:
+1 - was about to report this issue. It seems to be a problem on Filters/Logic, especially:

https://blueprintnotincluded.org/b/663cd22fcd58d2ec233f06c6

PR #39, opened 2024-05-09
Fix issue with Automation Wires being behind Automations

andrewvaughan:
I'm not a typescript guy, but this seemed like the proper fix for #34 - make sure to please test my work.


Sinetheta:
sorry I never replied to this. you're right and this will eventually fix the problem. I say eventually because it needs to get processed by our scripts to become a new database.json (and sprites). However when I ran those scripts I saw a number of untracked changes. For simplcity I'm just going to merge this as-is then work on the data scripts separately.

PR #109, opened 2026-07-08
feat(seed): expand dev seeds with full social graph + WorkOS dev login

Sinetheta:
## What & why

Reworks `npm run seed:dev-blueprints` from a fork-only fixture into a complete community-feature fixture, so the branch's social features (forks, likes, comments, follows, feedback) can be validated in the browser against real data.

Based on `feat/fork-blueprint-version` (not `master`): the `blueprint-version` model/service the seed forks depend on only exist on that branch. Merge this into the fork branch; it rides along to master when that does.

## What shipped

- **Categorization actually fires.** Placeholder blueprints are keyed by real prefab IDs and their `gameVersion` / `category` / `modded` are **derived** via `deriveGameVersion` / `deriveCategory` / `deriveModded` against `database-2024.json` — the same path production uses — rather than hand-set. Single-building probes cover each function category, plus deliberate edge cases:
  - bare `GasPump` → **Untagged** (ventilation/hvac is intentionally unmapped)
  - `Electrolyzer` + `BatteryModule` → **spacedOut**
  - unknown prefab → **modded=true**

  The run summary prints the derived result per blueprint, so a miscategorization is visible immediately.
- **Full social graph:** likes distributed across users; real forks (`BlueprintVersion` + `forkedFrom` + `forkCount`) including a **fork-of-a-fork** to exercise multi-hop lineage/version history; comments (top-level + reply that bumps `lastActivityAt`, an `@mention` and a `/b/<id>` reference run through `sanitizeCommentBody` into `{{user:…}}` / `{{blueprint:…}}` tokens, an edited comment, a soft-deleted comment); follows; and feedback (open + resolved) for the admin queue.
- **Dev login for a WorkOS site.** All login proxies to WorkOS, so DB-only seed accounts can't sign in through the browser — there is no password bypass. The script now mints a JWT per user with the same `generateJwt` + `JWT_SECRET` the real endpoint uses (`dev_admin` gets an **admin-role** token) and prints a `localStorage.setItem('blueprintnotincluded-token', …)` snippet. Paste into the console on the site origin, reload, and you're that user.
- **Idempotent:** cleanup now also removes seeded comments/follows/feedback, so re-runs stay a clean no-op.

## Verification

- `npx tsc --noEmit` clean.
- Ran `npm run seed:dev-blueprints` twice against the local dev DB: idempotent (5 users, 10 sources, 4 forks, 5 comments, 5 follows, 2 feedback each run).
- DB spot-check confirmed: reference tokens generated (`{{blueprint:…}}`, `{{user:…}}`), reply bumps parent `lastActivityAt`, fork-of-fork lineage present.
- All 10 derivation outcomes match expectations (each category tagged; Gas Pump & bare wire Untagged; Spaced Out probe = spacedOut; modded probe = modded=true).

## Notes / not included

- Placeholders are not visually realistic (1×1 thumbnail, minimal `data`) — intentional, per the goal of tripping metadata/social logic rather than rendering art.
- The unrelated in-flight `blueprint-controller.ts` edit on the working tree was left untouched (not part of this change).

### DIFF
--- a/app/api/batch/seed-dev-blueprints.ts
+++ b/app/api/batch/seed-dev-blueprints.ts
@@ -4,29 +4,34 @@
 // replies, edits, soft-deletes, @mentions and /b/ references), follows and feedback.
 //
 // Usage:
-//   npm run seed:dev-blueprints
+//   npm run seed:dev-blueprints   # full social-graph seed (destructive re-seed)
+//   npm run seed:dev-user         # ONLY ensure/restore the protected user + print its token
 //
-// Re-running is safe: it wipes everything it previously created (matched by the fixed
-// dev @bpni.local users below) before recreating it.
+// Re-running is safe: it wipes the disposable @bpni.local fixture and recreates it.
 //
-// --- Logging in as a seed user (WorkOS) ------------------------------------------
+// --- Logging in without WorkOS ---------------------------------------------------
 // The site authenticates through WorkOS (see agent/WORKOS_PLAN.md): a plain DB user
-// with a password CANNOT log in through the browser, because /api/auth/login proxies
-// to WorkOS and DB-only accounts come back as `legacy_account`. There is no dev
-// password bypass. Instead, this script mints a JWT for each seed user with the same
-// `generateJwt` + `JWT_SECRET` the real login endpoint uses, so the token validates
-// against the expressjwt middleware identically. The script prints a ready-to-paste
-// snippet per user; run it in the browser console (on the site origin) to "log in":
+// CANNOT log in through the browser, because /api/auth/login proxies to WorkOS and
+// DB-only accounts come back as `legacy_account`. There is no dev password bypass.
+// Instead this script MINTS a JWT signed with the same `JWT_SECRET` the real endpoint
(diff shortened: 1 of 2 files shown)


Sinetheta:
Follow-up pushed: **durable protected validation user** (`dev_you`) — addresses "I need one working user that survives DB resets without a WorkOS flow each time."

- Fixed deterministic `_id` (`d0d0…`) so a minted token keeps authenticating across any number of resets.
- Excluded from destructive cleanup — the full re-seed regenerates its *content* but never deletes the account.
- `npm run seed:dev-user` restores just this user + reprints its token (30-day, admin, alpha) with no WorkOS round-trip — one-command recovery after any drop.

Verified end-to-end against the running backend: minted token authenticates through the live `expressjwt` middleware (`GET /api/feed` → 200 with token, 401 without), `_id` stays stable across a full re-seed and a simulated drop+restore, re-runs idempotent.

Sinetheta:
Superseded by #110, which rebases the same two seed commits onto master now that feat/fork-blueprint-version has merged. Continuing there.

PR #118, opened 2026-07-09
fix(previews): load textures per blueprint; smoke-test renderer in CI

Sinetheta:
## What

Second prod failure after #117 shipped the canvas binding: the worker got through pixi-shim init, then the **whole DO component died** (`exited with code: 128`, 504s while restarting) ~5s into texture preload. The worker preloads all ~1,700 registered textures (~400MB RSS); parent API process + worker together exceed the instance's memory limit and the cgroup OOM kill takes down the container.

## Fix

**Selective texture loading.** The texture closure for one blueprint is statically enumerable after `importFromMdb`: `DrawPart.flatIconId` (flat icons + 16 connection-bitmask variants), `spriteModifier → SpriteInfo.imageId` (atlas-backed parts, element tiles), and utility-port marker sprites. The worker now decodes only those ids per render.

Measured against real prod data (top 5 most-liked blueprints from a prod dump, 117–502 items each): **213MB at ready, 250–340MB after five distinct blueprints**, zero placeholder warnings, renders pixel-identical (element paint, info icons, connectables all verified). Cold start drops ~10s → ~1s.

**RSS self-cap.** Textures accumulate across distinct blueprints, so a long-lived worker drifts back toward the full-preload footprint. Once RSS crosses `PREVIEW_WORKER_MAX_RSS_MB` (default 384) the worker exits between renders; the parent re-forks on the next request. A miss on `getNewBaseTexture` now warns + renders a placeholder instead of crashing the render.

**Diagnostics that were missing when prod failed:**
- parent logs worker exit `code`/`signal` (crash vs OOM kill vs idle shutdown)
- worker logs RSS at ready and after each render

**CI smoke test (the testing-strategy gap):** nothing ever exercised the render path inside the deploy image. The worker gains `--smoke`: renders a built-in fixture covering every texture path (flat icon, tileable tile, connectable wire, utility ports), fails on any missing texture or undersized output. `publish.yml` now runs it in the built image under a **512MB memory cap** before pushing to the registry — this step would have caught both the missing `canvas.node` and the OOM.

## Verification

- Smoke test passes locally (310KB PNG, rss=226MB)
- 5 real prod blueprints rendered through the worker over IPC, output visually verified
- Backend suite: 384 passing; 2 failures are the pre-existing `workos-provision` network flakes (fail on master too)

## Note for ops

If prod instance is 512MB, parent (~150–200MB) + worker cap (384MB) is still tight — consider 1GB, or lower `PREVIEW_WORKER_MAX_RSS_MB`. The new RSS logs will show the real envelope after deploy.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -59,6 +59,14 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
         
+      - name: Smoke test preview renderer
+        # Renders a fixture blueprint inside the deploy image under a
+        # prod-like memory cap: catches missing native bindings (canvas),
+        # broken asset layout and memory regressions before the push.
+        run: |
+          docker run --rm --memory=512m --memory-swap=512m bni-deploy \
+            node app/api/services/preview-render-worker.js --smoke
+
       - name: Verify version information
         run: |
           echo "Verifying version information in built image..."
--- a/app/api/services/preview-image-service.ts
+++ b/app/api/services/preview-image-service.ts
@@ -252,6 +252,9 @@ export class PreviewImageService {
 
       worker.on('exit', (code, signal) => {
         clearTimeout(startTimer);
+        // Stale exit: this worker was already stopped/replaced (kill() fires
+        // 'exit' asynchronously). Don't touch the current worker's state.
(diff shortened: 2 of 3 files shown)


Sinetheta:
@coderabbit review

PR #123, opened 2026-07-09
feat(previews): durable Mongo storage for preview images + backfill

Sinetheta:
## What

Phase 3 of the preview render performance & durability plan: rendered preview images now survive redeploys.

- **New `previewimages` collection** (`PreviewImageModel`): `{ blueprintId, variant, bytes, contentType, renderedAt, sourceModifiedAt }`, unique index on `{ blueprintId, variant }`. ~200KB × 3 variants × ~4,500 blueprints ≈ 1GB, included in normal backups.
- **Two-tier read path** in `PreviewImageService`: disk (L1) → Mongo (L2, hydrates disk on hit) → render. Renders write both tiers. Mongo freshness mirrors the disk mtime rule: a row is fresh while `sourceModifiedAt >= blueprint.modifiedAt`, so saves/forks/restores invalidate naturally.
- **Serving from Mongo is not rendering** — the L2 path stays live even when rendering is disabled (kill switch / worker failure), so a redeployed container serves the whole corpus without re-rendering anything.
- `prerender` (render-on-write) now also skips when the durable rows are fresh, not just the disk cache.
- **Backfill script**: `npm run backfill-previews` (and `:dry-run`) iterates all non-deleted blueprints through the same serialized worker queue, skipping fresh rows — rerunnable/resumable, per-25 progress logging, non-zero exit if any renders failed.

## Design decisions

- Store all three variants (not master-only) per the spec's default — simpler, and halving storage wasn't worth re-running sharp on every L2 miss.
- Mongo read/write failures degrade gracefully (log + fall through to render / disk-only) — durability is best-effort per render; the backfill or next render retries.
- Soft-deleting a blueprint leaves its rows in place (invisible, harmless) — cleanup can piggyback on whatever purges soft-deleted blueprints later, per spec.
- `PreviewController.loadRenderData` made public so the backfill feeds the queue with the same currentVersionId-aware data resolution as the live read path.

## Verification

- `npm run tsc` clean; backend suite: 388 passing, 0 new failures (the 12 failing WorkOS suites fail identically on master locally — they time out without network access to WorkOS).
- 5 new tests: durable rows written on render, L2 hit serves without render + hydrates disk, stale rows re-render (upsert, no duplicates), disabled-rendering redeploy acceptance, prerender skip.
- Dry-run against the local bpni-prod dump: 4,484 blueprints, all reported stale, nothing written.
- Full backfill against the local bpni-prod dump is running now (~2–4s/blueprint through the real worker); will report the final fresh/rendered/failed counts in a comment before merge.

## Note for ops

- This grows prod dumps by ~1GB; `mongodump` remains fine.
- After deploy, run `npm run backfill-previews` once (DO app console) to populate the corpus; subsequent redeploys then serve every existing blueprint with zero re-renders.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -90,6 +90,7 @@ and their `npm run generate*` / `seed` / `enhancedSeed` / `testCanvas` entries n
 Remaining batch utilities:
 - `npm run fixHtmlLabels` - Fix HTML formatting in labels.
 - `npm run derive-metadata` - Backfill `gameVersion` and `modded` on all blueprint documents from stored building IDs. Use `--dry-run` flag (`npm run derive-metadata:dry-run`) to preview counts without writing.
+- `npm run backfill-previews` - Render preview images for all non-deleted blueprints and store them durably in Mongo (`previewimages` collection). Skips blueprints whose durable rows are already fresh, so it's rerunnable/resumable. Use `--dry-run` (`npm run backfill-previews:dry-run`) to report the fresh/stale split without rendering or writing.
 
 ### Docker
 - `docker-compose up` - Full development environment with database
--- a/__tests__/api/preview.test.ts
+++ b/__tests__/api/preview.test.ts
@@ -13,7 +13,11 @@ process.env.NODE_ENV = 'test';
 import { TestSetup } from '../setup/testSetup';
 import { BlueprintModel } from '../../app/api/models/blueprint';
 import { BlueprintVersionModel } from '../../app/api/models/blueprint-version';
-import { PreviewImageService } from '../../app/api/services/preview-image-service';
+import { PreviewImageModel } from '../../app/api/models/preview-image';
+import {
+  PreviewImageService,
+  PREVIEW_VARIANTS,
+} from '../../app/api/services/preview-image-service';
 import { Types } from 'mongoose';
 
 // Waits for a fire-and-forget prerender to finish (or fail) by polling.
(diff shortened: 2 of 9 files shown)


Sinetheta:
## Verification results (local bpni-prod dump)

Ran the backfill against the 4,484-blueprint prod dump; stopped deliberately at 1,550 (34%) — the remaining renders add no verification value since prod populates itself (see below).

- **1,480 rendered + 69 skipped-as-fresh + 1 failed** in ~50 min (~1.9s/blueprint on the dev Mac)
- Resume/skip logic proven end-to-end: an interrupted run (the dev machine ran out of disk mid-run, killing local Mongo) restarted and skipped all previously stored rows in seconds, then continued
- The 1 failure is a modded blueprint referencing a building absent from the current database (`CosmicResearchCenter`); it logs, counts, and falls back to the legacy thumbnail — expect a handful of these on prod
- Redeploy acceptance verified by test: fresh Mongo rows serve with rendering disabled entirely

## Corrected storage estimate

Measured **~775KB per blueprint** (3 variants, BSON), not the spec's ~600KB — and WiredTiger compression doesn't help (webp/png bytes are already compressed). Full corpus: **~3.5GB, not ~1GB**. mongodump/backups grow by the same. Still well within cluster capacity, but worth knowing before merge; revisit R2/Spaces only if this becomes a problem, per the original decision.

## Prod rollout (unchanged)

Local results are not uploaded to prod (local machines only hold the read-only prod URI, by design). After deploy: `npm run backfill-previews` once in the DO app console — rerunnable/resumable, so an interrupted console session is harmless. At prod's ~3–5s/blueprint that's one overnight run; live traffic during it just falls back to legacy thumbnails when the render queue is busy.

Sinetheta:
Storage decision (Kevin, 2026-07-09): keep **all three variants** in Mongo (~3.5GB, ~23% of the prod cluster's 15GB quota). Measured per-variant split: card 53KB / hero 177KB / og.png 519KB avg. Master-only storage (~0.8GB, deriving card/og on Mongo hits) was considered and declined for read-path simplicity — revisit only if the corpus outgrows the quota.

PR #130, opened 2026-07-11
feat(forks): attribute editor copy-saves as forks

Sinetheta:
## What ships

The editor's old "copy" flow — a non-owner opens someone else's blueprint and saves it — used to silently create a detached copy with no link to the original. That copy is now attributed as a fork, using the same top-level fork concept as the fork button:

- **Frontend** (`blueprint-service.ts`): `saveBlueprint` includes `sourceBlueprintId` (the id the editor was opened from) in the upload payload whenever one exists. The server decides whether it matters.
- **Backend** (`blueprint-controller.ts`): on the create path of `POST /api/uploadblueprint`, if the requester can view the source and doesn't own it, the new document gets `forkedFrom` (source id + current version id, via `ensureCurrentVersion`), the source's `forkCount` is incremented, and its owner receives the existing `fork` notification — mirroring `POST /api/blueprints/:id/fork`.

## Design decisions

- **Best-effort attribution, never a save failure**: a malformed id is treated as absent; a missing/deleted/unviewable/self-owned source just skips attribution. The old copy behavior is fully preserved — same results, better attribution.
- **Create path only**: overwriting your own document (including the resurrect-soft-deleted path) never gains attribution; a copy's later overwrite saves keep their original `forkedFrom` and don't re-increment `forkCount`.
- **Name is kept**: unlike the fork button, no " fork" suffix — the user typed the name in the save dialog.
- **No self-forks**: owner saving their own blueprint under a new name stays an unattributed copy, per the feature's scope (non-owner saves).

## Verification

- Backend: full Mocha suite 447 passing (incl. 5 new copy-as-fork tests covering attribution, owner/self skip, unknown/malformed source, draft source, overwrite idempotence). One pre-existing comments-API timeout flaked in the full run and passes in isolation.
- Frontend: 679 passing (2 new specs for `sourceBlueprintId` presence/absence in the save payload).
- `npm run tsc` clean.

### DIFF
--- a/__tests__/api/blueprint-forks.test.ts
+++ b/__tests__/api/blueprint-forks.test.ts
@@ -529,7 +529,7 @@ describe('Fork + BlueprintVersion API', function () {
       const token = testData.users.user2.generateJwt();
       const response = await upload(token, {
         name: 'Copy Of Hidden Draft',
-        sourceBlueprintId: draft._id.toString(),
+        sourceBlueprintId: String(draft._id),
       });
       expect(response.status).to.equal(200);
       await settle();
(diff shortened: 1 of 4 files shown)


Sinetheta:
@coderabbit review

PR #145, opened 2026-07-12
chore: ESLint 9 flat config migration + Node toolchain alignment

Sinetheta:
## Why

CodeRabbit's ESLint runs against this repo were failing with:

```
NOT SUPPORTED: option missingRefs. Pass empty schema with $id that should be ignored to ajv.addSchema.
TypeError: Cannot set properties of undefined (setting 'defaultMeta')
    at ajvOrig (/frontend/node_modules/@eslint/eslintrc/dist/eslintrc.cjs:1626:27)
```

Root cause: ESLint 8's legacy-config loader (`@eslint/eslintrc`) requires ajv v6, but the frontend tree hoists ajv v8 (via `@angular-devkit`). Locally npm nests an ajv v6 copy under `@eslint/eslintrc` so it works; CodeRabbit's installer resolves the hoisted v8 instead and the loader crashes. Migrating to ESLint 9 flat config removes `@eslint/eslintrc` from the load path entirely, so a bare `eslint <file>` works regardless of hoisting.

## What shipped

**Frontend lint toolchain**
- ESLint 8 → 9, `@typescript-eslint/*` v7 → `typescript-eslint` v8, five `@angular-eslint/*` packages → `angular-eslint` v20 meta package (v7 of typescript-eslint was below angular-eslint 20's supported range), `eslint-config-prettier` 8 → 10
- `.eslintrc.json` → `eslint.config.js` (flat config), preserving all prior rules and adding the `eslint/js` + `typescript-eslint` recommended sets
- `no-explicit-any` off (351 pre-existing hits — separate burn-down if ever desired); `no-unused-vars` honors the codebase's `_`-prefix convention; vendored `src/app/primeng-changes/` ignored
- No type-aware rules are in use, so no project service — bare `eslint` runs need no tsconfig program membership (this also fixed 93 "file not found by project service" errors for files outside the `main.ts` import graph)

**Violations fixed (was 448, now 0)**
- `ng lint --fix` autofixes: ~270 `prefer-const` / `no-var`
- `sanitize-filename` `require()` → proper default import (typechecked)
- Dead code removed: `if (false)` branch in tool-service, `1; // required for type` + empty `ngOnInit` in like-widget, unused `declare let PIXI` in draw-room-overlay
- `custom-event-manager.ts` keeps `Function` types with a scoped disable — signature must match Angular's `EventManager` API

**Dead weight removed**
- Protractor e2e: the `@angular-devkit/build-angular:protractor` builder no longer exists in Angular 20, so `ng e2e` has been broken since that upgrade. Removed the target, `e2e/` sources, and `protractor`/`ts-node` devDependencies
- Root `tslint.json` (TSLint retired years ago)

**Node toolchain alignment**
- Backend `@types/node` 14 → 20 (runtime is Node 20)
- `publish.yml` node-version 20.18.0 → 20.19.4 (matches engines `^20.19` and the other two workflows)
- `tsconfig.admin.json` now referenced from the solution tsconfig with a widened include, so `src-admin` gets language-service coverage

## Verification

- `ng lint` green for both projects; bare `npx eslint .` and `npx eslint <the file CodeRabbit choked on>` exit 0
- `tsc -p tsconfig.app.json --noEmit` clean; backend `tsc -b` clean
- Frontend: 750 tests passed, 2 skipped (Vitest)
- Husky pre-commit (lint-staged → ng-lint-staged → prettier → tsc) ran clean on the migration commit itself

## Deferred

- Frontend still on prettier 2 (backend on 3) — unifying would reformat broadly, kept out of scope
- Backend/`lib` have no ESLint coverage; can be added later with the same flat-config pattern
- `no-explicit-any` burn-down (351 sites)

## CI restructure (second commit)

Lint, typecheck, test, and build now run as **parallel jobs** instead of one serial pipeline:

- **Frontend CI**: `lint` / `typecheck` / `test` / `build`. The lint job installs only frontend deps (no type-aware rules → no lib build), so it's the fastest signal. Wall-clock is now the slowest job, not the sum.
- **Frontend typecheck was a no-op**: `frontend/tsconfig.json` is solution-style (`files: []`), so the old `npx tsc --noEmit` step checked *nothing*. CI now checks the app, spec, and admin projects directly; the husky pre-commit hook got the same fix (app project only, for speed).
- **Backend CI**: `typecheck` (no mongo service) runs parallel to `test`. `tsc -b` builds `lib/` via project reference, so redundant `build:lib` steps were dropped here and in publish's quick-check.
- Removed stale `jest.config.js` path filter (Mocha repo — now `.mocharc.json` + `migrations/**`) and the backend artifact upload that never had files to upload.


Sinetheta:
@coderabbit review

PR #153, opened 2026-07-17
feat(discover): show active filters above the grid, not just the sidebar

Sinetheta:
## Summary

Found while investigating a report that "ranching category is empty": it wasn't a detection bug (ranching has 499 tagged blueprints on prod, second-largest category). The actual URL was `?category=ranching&rooms=latrine&sort=popular` — a nonsensical combination (ranching buildings never form a latrine), so zero results is correct. The problem is UX: nothing on the page explains *why* it's empty.

- The only record of active filters lived in the sidebar, which scrolls out of view on longer pages.
- The empty state was a generic "No Results" card, indistinguishable from an actual outage.

## Change

- Added a chip bar pinned above the results grid, in `<main>` rather than the sidebar, so it's visible regardless of scroll position. Renders one removable chip per active facet: search term, category, subcategory, gameVersion, modded, and each active room — plus a "Clear all" action.
- Folded the old `filterModded` banner sentence ("Showing modded blueprints only.") into this bar as a chip. Left the `forkedFrom` banner alone — it has no sidebar control and would just render an opaque id as a chip.
- The empty-state message now distinguishes "No blueprints match these filters" (filters are active) from the cold-start "No Results" (nothing active) — same sentinel-card mechanism, just a different string picked at push time.
- Room chips reuse the existing `roomTypeLabel()` helper for human-readable labels (e.g. "Latrine"); category/subcategory/gameVersion chips stay raw enum strings, matching the existing card-badge convention.
- Removing any filter chip clears just that facet and refetches; rooms is effectively single-select today (the sidebar replaces rather than appends), so any room chip's remove clears the whole rooms filter — matches the sidebar's own semantics.

## Test plan
- [x] `npx tsc --noEmit` (frontend) — clean
- [x] `eslint` on the touched files — clean
- [x] `browse-page.component.spec.ts` — 56/56 passing (14 new: chip building/labels/removal, no-op guards, empty-state message selection)
- [x] Live-verified against the local dev server with Playwright: navigated to `?category=ranching&rooms=latrine&sort=popular`, confirmed both chips render with correct `aria-label`s and the filtered empty message shows; clicked "Remove filter: Latrine", confirmed the URL drops just `rooms`, the chip bar updates to show only `ranching`, and a real result reappears

### DIFF
--- a/frontend/src/app/module-blueprint/components/browse-page/browse-page.component.css
+++ b/frontend/src/app/module-blueprint/components/browse-page/browse-page.component.css
@@ -128,6 +128,46 @@
   font-size: 0.9rem;
 }
 
+/* Active-filter summary, pinned above the grid so it stays visible even
+ * when the facet sidebar is scrolled out of view. */
+.active-filters-bar {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+  gap: 8px;
+  margin: 0 0 16px;
+}
+
+.active-filter-chip {
+  display: inline-flex;
+  align-items: center;
+  gap: 6px;
+  font-family: var(--p-font-family);
+  font-weight: 500;
+  font-size: 12px;
+  letter-spacing: 0.01em;
+  padding: 3px 8px 3px 9px;
+  border-radius: 2px;
+  background: var(--bni-chip);
+  color: var(--bni-body);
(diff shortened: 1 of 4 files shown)


Sinetheta:
@coderabbit review

PR #160, opened 2026-07-18
feat(avatars): attach in-game hats sheet to generation prompts (v3)

Sinetheta:
## What shipped

Avatar generation now attaches a second committed reference image — `assets/avatar-reference/duplicant-hats-sheet.jpg`, a sheet of ~34 real ONI hats (mining helmets, ranger hats with teeth, medic caps, VR visors, space helmets, …) — alongside the duplicant style sheet on every Gemini call.

- **Prompt templates bumped to v3** (`duplicant-grid-v3`, `face-duplicant-grid-v3`; ids persisted per avatar/batch as before).
- **Hat rule:** exactly one of the four grid characters is bare-headed; the other three each wear a different hat. Per-character briefs are randomized from a list that mostly names hats actually on the sheet (so they come out recognizable, exact or riffed) plus two "invent your own hat in this style" entries.
- **Face flow:** the user photo shifts to the *third* attachment slot; the same one-hatless/three-hatted rule applies across the four portraits of the person.
- **Fail closed on either sheet:** prompts refer to attachments positionally, so a missing sheet would silently mislabel every other attachment — `generateBatch` refuses instead (extends the existing style-sheet fail-closed behavior). New env override `AVATAR_HATS_REFERENCE` mirrors `AVATAR_STYLE_REFERENCE`.
- Also refreshes the committed style sheet jpeg and updates `agent/AVATARS.md`.

## Design decisions

- Hat briefs name sheet hats descriptively ("the ranger hat with a band of teeth from the hats sheet") rather than by grid position — the model matches drawings by appearance more reliably than by row/column.
- `deploy.Dockerfile` already copies the whole `assets/avatar-reference/` dir, so no deploy changes needed.

## Verification

- `npm run tsc` clean.
- Avatar suite: 36/36 passing, including new specs for the exactly-one-hatless invariant across rng seeds, both sheets attached as the first two references, three references on the face path, and fail-closed on a missing hats sheet.
- No real generation run (costs money) — recommend `npm run avatars:smoke` after merge to eyeball hat quality.


Sinetheta:
@coderabbit review

PR #171, opened 2026-07-19
Update asset hitbox

Sinetheta:
I have update the mod to better size ALL exported images this results in an unfortunate amount of churn to most images but should be a one-time only penalty as we now de-dupe images very well.


Sinetheta:
@coderabbit ignore

PR #176, opened 2026-07-25
feat(import): carry element mass and temperature defaults

Sinetheta:
Imports the new element defaults from the game export so the element-notes feature has real per-element mass and temperature data to work with.

## What the converter now gathers

`elements.json` ships three new fields; this carries them (plus `lowTemp`/`highTemp`, which bound the temperature picker) through `convert-export-2024.ts` into `BuildableElement`:

| field | unit | notes |
|---|---|---|
| `maxMass` | kg | sim cell capacity |
| `defaultMass` | kg | what the game pre-fills |
| `defaultTemperature` | Kelvin | same scale as `lowTemp`/`highTemp` |
| `lowTemp` / `highTemp` | Kelvin | phase-change bounds |

The website can now seed a picker per element instead of applying one hardcoded constant to all of them (`BlueprintItemElement.defaultMass = 0`, `BlueprintItem.defaultTemperature = 30°C`).

These can't be hardcoded or derived. Crude Oil defaults to 350 K inside a 233–673 K range; Chlorine defaults to 600 kg against a 1000 kg capacity. And `gas.yaml` ships without any of them — the game applies `maxMass 1.8` / `defaultMass 1.0` to every gas at load time, so a static table would bake in constants that break when Klei changes them.

## `state` normalization

The export writes `state` as the enum **name** for most solids (`"Solid"`) but its raw **numeric** value elsewhere (`"5"`, `"6"`, `"20"`), and the numeric form carries flag bits above the phase. `parseElementState()` accepts either and masks to the low 2 bits, feeding a new `ElementState` enum (Vacuum/Gas/Liquid/Solid) — same class of quirk the converter already handles for `ConnectionType`.

Result across all 212 elements: **32 gas, 52 liquid, 125 solid, 3 vacuum**. The 32 gases match the count the export contract documents.

## Validation

The import now fails on either half of the export's own contract:

- **every gas has `maxMass == 1.8` and `defaultMass == 1.0`** — a violation means the exporter read `gas.yaml` instead of the runtime `Element`, and every gas mass picker is silently wrong;
- **`defaultMass <= maxMass`** for every element, which the game enforces itself in `ElementLoader.CheckElements()`.

Both are also asserted against the committed database in `database-validation.test.ts`, along with the documented value table (Water/Ice/CrudeOil/Chlorine/Oxygen/Steam) and the state normalization — so a bad reimport fails in CI, not just at the console.

## ⚠️ This export is also a newer game build

`U59-740622-SCRPAN`, up from `U59-737790`. It brings **14 new buildings** (473 → 487) and 97 changed icons alongside the element data. Splitting that out would mean hand-editing generated JSON and the next import would reintroduce it, so it ships here as its own commit.

Those 14 buildings are all from **two DLC ids the analyzer doesn't know** (10 × `DLC3_ID`, 4 × `DLC4_ID`), which surfaced a **pre-existing bug**: `DLC5_ID` is mapped to `bionicBooster`, but the Bionic Booster content (electrobanks, gunk, robo-pilot) is `DLC3_ID` — `DLC5_ID` is the aquatic pack (UnderwaterRanchStation, ReefGenerator, SushiBar…). 15 buildings are mislabelled, and `gameVersion` is stored per blueprint, so prod data is already affected.

**Deliberately not fixed here.** It needs two new `GAME_VERSIONS` values (stored strings, so naming them wrong costs a backfill), frontend facet updates, and a `derive-metadata` re-run. Written up in `agent/TODO.md` + `spec/dlc-game-version-mapping.md`. The contract test carries an explicit `pendingDlcIds` allowlist meanwhile, so it still catches any *other* new DLC id.

## Verified

- `npm run import:2024` clean — zero issues across every validation line.
- Backend **663 passing**, frontend **985 passing**, `npm run tsc` clean.
- Spot-checked the emitted record: `Oxygen` → `state: 1, maxMass: 1.8, defaultMass: 1, defaultTemperature: 300, lowTemp: 90.19, highTemp: 10000`.

## Not included

No UI. This is the data layer for the element-notes feature; nothing consumes the new fields client-side yet.


Sinetheta:
@coderabbit review

PR #191, opened 2026-07-31
Converge website Info annotations on the mod's world notes

Sinetheta:
## The bug

Downloading a website blueprint that shows info notes, then re-uploading it, produces a blueprint with no notes. Reported against [Hydrogen Vent Tamer](https://blueprintnotincluded.org/browse/6a43fc3390f889ca0adef97b) — 15 notes on the site, zero in the file.

The site had **two** annotation models:

| | legacy `Info` | mod `worldNotes` |
|---|---|---|
| stored as | a `blueprintItem` with id `Info` | `data.worldNotes[]` |
| exports to `.blueprint` | **never** | yes |
| renders in server previews | **yes** (it's a building) | **no** |

Exactly inverted — each covered the gap the other had. `toBniBlueprint` skipped `Info` items because there is nowhere in the BlueprintsV2 format for a building the game cannot build, so every website annotation was dropped on download. A blueprint made *in game* round-trips fine, which is why the bug looked intermittent: those carry real world notes, and the server serves their raw upload byte-exact without re-serializing.

## What shipped

**The mod is the source of truth, so world notes win.** `Info` survives as an input format only.

- `lib/src/blueprint/note-conversion.ts` — `infoBuildingToWorldNote`, applied in `Blueprint.importFromMdb`. Stored `Info` items become world notes on read and are never written back, so a document is correct whether or not the migration has reached it, and any re-save converges it.
- The `!= infoId` guard in `toBniBlueprint` is gone (nothing with that id reaches the loop any more). Notes export with `blueprintVersion: 3`; `modSanitizeOffset` already re-origins world notes alongside buildings, so converted annotations stay glued to what they annotate.
- Legacy editor UI removed: the build-menu info button, the `info-input` / `info-input-icon` components, the `item-collection-info` info branch, `BlueprintItemInfo`, `OniItem.isInfo` and the synthetic `Info` OniItem. The Notes tool is now the only way to annotate.
- `lib/src/drawing/note-markers.ts` — the pure marker rules (sprite, colour, size, symbol table) moved out of the frontend so the backend can share them. The editor overlay keeps only the per-frame work: texture pooling and the selection ring.
- The preview worker draws world notes. Without this, converging the models would have blanked annotations off ~2,000 blueprint cards, since the worker walks `blueprintItems` and nothing else.
- `migrations/20260730000000_info-items-to-world-notes.js` — moves stored `Info` items into `data.worldNotes` across `blueprints` and `blueprintversions`.

## Decisions

**`frontColor` is dropped.** The website badge had a fill colour and a glyph colour; a world note marker is a single-colour sprite, so the second has nowhere to go. 92 distinct values were in use, so this is real (if minor) loss — the alternative was a `bni/notes` metadata sidecar, judged not worth a second coordinate-sync surface for one colour field.

**Badge colour and icon are written explicitly**, not left to the world note defaults. The two models' defaults differ (`#007ad9` vs `#3b82f6`), so an implicit conversion would silently repaint every unstyled note.

**Migration `down` is exact for what `up` converted, approximate afterwards.** A note authored in game and imported later also converts back to `Info` — its symbol and tint map cleanly, so it renders correctly, but it is no longer the note its source file carried. Two prod documents are in that category. Rollback is a break-glass operation in the deploy window; this is the honest behaviour for that moment.

**Terrain annotations still do not appear in server previews.** Same pre-existing gap, out of scope here.

## Verification

- Backend **796 passing** (10 new: conversion mapping, import behaviour, export round trip, sanitize offset, coexistence with native notes)
- Frontend **1120 passing**, lint clean, `tsc` clean on all three trees
- Round-tripped the real reported blueprint through the compiled lib: 67 buildings + **15 notes** out, 15 notes back in on re-import
- Rendered it through the actual preview worker — all 15 markers present, in position, in their original colours
- Migration run against a **full prod restore** (5,271 blueprints): 17,441 annotations across 2,017 documents (38% of the corpus) plus 474 across 66 versions; `down` reverted cleanly; `up` re-run is a no-op

## Deploy

Run `npm run migrate:up` from `/bpni/build` after deploy (back up first). Nothing breaks if it is delayed — read-time conversion covers the corpus meanwhile; the migration is what makes the *stored* data single-model.


Sinetheta:
Both review findings verified and fixed in a895466d.

**`down` reversed notes it did not create** — valid, and not hypothetical: the original prod rehearsal showed it, with a down/up cycle reporting 17,453 conversions against the first pass's 17,441. The extra 12 were the two documents whose world notes came from the game.

`up` now records the `Info` items it removes (with their array positions) in a `migration_20260730_info_notes` collection, and `down` restores exactly those. Provenance lives in a side collection rather than as a marker field on the notes — a field would be permanent pollution of the blueprint model, written into every exported `.blueprint` from then on, to serve a rollback expected never to run.

**Cursor snapshots could clobber concurrent saves** — valid. Every write now carries the array state it was computed from as a compare-and-set filter; a document that changed underneath fails to match and is re-read on a later pass. Batching is unchanged. Retried documents are counted once, so a contended run cannot report more conversions than the corpus holds.

While fixing the first, the reversal became exact rather than merely equivalent: `Info` items are spliced back at their original indices (array order is the tie-break for equal z-indexes at render time), and an emptied `worldNotes` is unset rather than left as `[]`, matching what `toMdbBlueprint` writes.

### Verification (restored prod copy, 5,271 blueprints)

| check | result |
|---|---|
| `up` + `down` vs. real prod | **2,017/2,017** blueprints and **66/66** versions byte-identical |
| `up` re-run | no-op |
| `down` re-run | no records, no-op — cannot re-apply `Info` items |
| CAS semantics | stale write rejected, fresh write accepted, absent-field branch rejected |
| 4,160 concurrent saves during the run | every save preserved; end state identical to an uncontended run (17,453 annotations, 0 `Info` remaining) |
| rollback with mixed ownership | restored only the 1,977 documents the migration converted, leaving 40 the app had converted and 2 native ones alone |

One incidental finding from the race harness: an early version replayed a pre-migration snapshot verbatim, which reintroduced `Info` items after conversion. No app path can do that (`importFromMdb` converts on read), but it was useful to see the closing assertion refuse to report success rather than finish silently.

Sinetheta:
Second review round: one finding fixed (52d55b1f), one skipped.

### Fixed — `after` ran regardless of whether its own write landed

Valid, and the `down` direction made it a data-safety bug rather than a bookkeeping one. `bulkWrite` reports matches only in aggregate, so every callback in a flushed batch ran even when some compare-and-set filters were rejected. In `down` that callback deletes the provenance record — so a document whose CAS failed lost the record that makes it revertible at all, silently, and permanently once the run gave up retrying.

Updates are now issued individually in bounded parallel waves (50 in flight), so each callback sees its own document's `matchedCount`. Round trips stay in the same order of magnitude as batching gave us — a full `up` over the prod copy runs in ~12s.

Conflict tracking got more precise as a side effect: a contended pass now re-reads the documents that actually changed rather than every document that shared a batch with one, and the reported totals count writes that landed rather than writes that were planned.

**Verification.** Ran the real migration against a synthetic 400-document database while one document was edited continuously so its CAS could never match:

```
blueprints: 1 document(s) changed during the pass, re-reading   (x3 passes)
down threw: yes — still contended after 3 passes
victim kept its provenance row: true
provenance rows left: 1 (only the victim)
documents reverted to Info: 399
```

Under the previous code all 400 records would have been deleted, including the one document that was never restored. Re-checked on the prod copy too: up + down still leaves all 2,083 affected documents byte-identical to prod, and a run under concurrent saves still converges on the same state as an uncontended one (17,453 annotations, 0 `Info`, all 40 concurrent saves preserved).

### Skipped — moving the `$unset` out of the CAS update

The combined `$set` + `$unset` is a single-document update, so it is already atomic, and the CAS filter pins **both** arrays — a document that gained a note concurrently fails the match rather than having it unset. Confirmed directly:

```
stale $set+$unset matched: 0        (rejected)
concurrent note survived: true
fresh matched: 1 | worldNotes key present: false | sibling data.* kept
```

The suggested replacement would be a step backwards: an `updateMany` over `data.worldNotes: {$size: 0}` carries no compare-and-set at all and is not scoped to this migration's own documents, so it would also strip empty arrays off documents the rollback never touched. Leaving it as-is.

PR #210, opened 2026-08-23
feat(search): Gemini romanized-Vietnamese title gate

Sinetheta:
## What ships

Blueprint titles like `Dien phan full` are Vietnamese typed without diacritics. Our own detector can't place them and Google mistranslates them outright, so they've been unreachable to an English searcher no matter how often the existing passes re-run. This adds a Gemini gate that identifies, restores and translates them, in front of the existing Google provider-side detection.

New: `gemini-vietnamese-title-provider.ts`, `vietnamese-title-prompts.ts`, `vietnamese-title-translation-service.ts`, migration `20260805000000_translation-unit-modes.js`. Wired into `derive-search.ts` and `search-index-service.ts`. Also closes out the offline A/B experiment that motivated the design (local-only tooling, never wired into the app).

## Why it refuses more than it accepts

A result is taken only when **all** of these hold: the restored Vietnamese strips mechanically back to the *exact* input (the anti-hallucination guard), the English differs from the input by more than punctuation and case, numbers and version markers survive, and the text passes the shared blueprint-name policy. Ambiguous answers and faithful echoes of English titles are discarded and **never cached**, so a re-run can't accumulate a bad answer.

The downside is bounded independently of any of that: `titleOriginal` keeps the authored text in the index, so a wrong translation can only ever *add* a spurious match, never remove a correct one.

## Spend guardrails

Fail-closed on three independent switches — the kill switch (`GEMINI_VI_TITLE_TRANSLATION_ENABLED`, literal `'true'`), the shared `GEMINI_API_KEY`, and a positive monthly micro-USD allowance defaulting to `0`. Token caps are code-pinned; env vars can only *lower* them. Budget is reserved at the per-call maximum before dispatch and never released back down, so a provider failure costs its reservation rather than going unrecorded; observed spend is tracked separately.

**Sizing note:** because reservations are ceilings, the allowance bounds *calls*, not dollars — each reserves 2,176 micro-USD regardless of batch size.

## Review findings — two defects found and fixed

Neither was caught by the existing tests.

1. **The Google detection pass silently stopped running when the gate was off.** With the kill switch unset — the default, and exactly the state prod deploys in — every candidate came back `invalid`, nothing reached `googleCandidates`, and the Google pass that ships on master did nothing at all, while logging `Gemini accepted 0/N` as though the gate had merely declined. Root cause: `invalid` conflated "the gate judged this and refused" with "the gate is switched off". Fixed by exporting `isVietnameseTitleGateActive()` (one source of truth, shared with the service's own spend check) and branching on it in both callers. While the gate *is* active its refusals stay final — re-asking Google about an `ambiguous` title is the guess the gate exists to prevent.
2. **`.env.test` never blanked `GEMINI_API_KEY`**, so the batch scripts' bare `dotenv.config()` would load a real key into the test process. Same leak this repo has been burned by before.

Also hardened: bulk updates now filter on the `sourceHash` they were read at, so a title re-authored mid-run isn't overwritten with a translation of the old text.

## The A/B review, now scored

The blinded native-speaker review sheet from the offline experiment had never been scored. It is now — all 48 rows, arms unblinded only after grading. Substantive cases only (10 of 12; English controls excluded):

| Arm | Provider | EN-acceptable | Unflagged wrong assertions |
| --- | --- | ---: | ---: |
| B `llm-end-to-end` (**shipped**) | Gemini restores + translates | **10/10** | 0 |
| D `restore-google` | Gemini restores, Google translates | **10/10** | 0 |
| C `google-vi` | Google, source forced `vi` | 2/10 | 3 |
| A `google-auto` | Google, auto-detect | 1/10 | 3 |

Scored by Claude, **not** by a native speaker — recorded honestly in the sheet. The fixture is elementary domain vocabulary and the arms separate by a margin far wider than the judgment involved (Google returns `Bơm khí tự động` → "Automatic gas bomb", `Kho chứa khí 2` → "Sourdough when 2"). Ordering is settled; individual scores are ±1.

**The specific worry going in was inverted by the data.** The concern was that the gate would silently resolve `ma` → `mã` with no ambiguity flag. It doesn't: on both deliberately ambiguous fixtures it returned status `ambiguous` with alternatives, and production rejects any non-`translated` status. The unflagged confident answers on those rows are *Google's* — `ma` → "and", `ma` → "ghost". The behaviour worth worrying about belongs to the arm this change replaces.

Replayed through the real accept rules: **8 accepted and correct, 2 refused as ambiguous, 2 refused as English echoes, 0 wrong acceptances.** That replay is now a test, so future drift in the accept rules fails against the reviewed run instead of surfacing in prod.

## Verification

- `npm run tsc` clean.
- Backend suite **1067 passing / 0 failing** (1064 baseline verified independently on a clean run before any changes — the prior report of a flaky avatar timeout did not reproduce; +3 new tests).
- Migration reviewed for idempotency in both directions; sets `mode` before touching the index, and `down` deletes vi-mode rows before restoring the narrower unique key.
- No network in tests: fakes for both providers, all keys blanked in `.env.test`.

## Rollout (ops, after merge)

Deploy has the gate off. Order matters:

1. DB backup, then `npm run migrate:up`.
2. `npm run derive-search:dry-run` — constructs no Gemini client, reports the exact census and maximum reservation.
3. Set `GEMINI_VI_TITLE_TRANSLATION_ENABLED=true` and an allowance chosen from that maximum, restart, then `npm run derive-search` once.

Rollback is switch-off plus allowance-zero; accepted cache rows stay usable and `Blueprint.name` is never modified.

## Known, deliberate

Only *accepted* translations are cached, so a re-run re-asks Gemini about every ambiguous and not-Vietnamese title. Bounded and documented; it's why a routine re-run isn't free.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -17,6 +17,9 @@ Dockerfile
 /frontend/tmp
 /frontend/out-tsc
 
+# Development-only, offline provider experiment; never compile it into a deploy image.
+/app/api/experiments
+
 # Only exists if Bazel was run
 /frontend/bazel-out
 
--- a/.env.sample
+++ b/.env.sample
@@ -22,7 +22,8 @@ WORKOS_CLIENT_ID="client_your_client_id_here"
 WORKOS_PLATFORM_ORG_ID="org_your_org_id_here"
 SITE_URL="http://localhost:3000"
 
-# Gemini avatar generation (agent/AVATARS.md)
+# Gemini features (avatar generation and the separately gated Vietnamese-title
+# search derivation)
 # Create a key at https://aistudio.google.com/apikey — image models need a
 # billing-enabled project (no free tier). Leave unset to disable the feature.
 GEMINI_API_KEY=""
@@ -31,6 +32,17 @@ GEMINI_API_KEY=""
 # AVATAR_CLASSIFY_MODEL="gemini-3.5-flash"
 # AVATAR_POOL_LOW_WATER="5"
 # AVATAR_POOL_REFILL="5"
+
+# Romanized-Vietnamese blueprint titles -> English search titles. Cache reads
+# remain available when disabled. New calls require all three: the exact true
+# kill switch, the shared GEMINI_API_KEY, and a positive integer micro-USD
+# monthly allowance. Keep false/0 until migrations and the no-network
+# derive-search dry-run have been reviewed.
+GEMINI_VI_TITLE_TRANSLATION_ENABLED="false"
+GEMINI_VI_TITLE_MONTHLY_BUDGET_MICRO_USD="0"
+# Hard caps may be lowered but not raised above these code-pinned defaults.
+# GEMINI_VI_TITLE_MAX_INPUT_TOKENS="4096"
+# GEMINI_VI_TITLE_MAX_OUTPUT_TOKENS="768"
 
 # User content translation (spec/user-content-translation-impl.md)
 # API key at https://console.cloud.google.com/apis/credentials, project needs
--- a/.env.test
+++ b/.env.test
@@ -11,3 +11,8 @@ FRONTEND_URL=http://localhost:4200
 WORKOS_API_KEY=sk_test_dummy_key_for_testing
 WORKOS_CLIENT_ID=client_test_dummy_client_id
 GOOGLE_TRANSLATE_API_KEY=
+# Batch scripts call a bare dotenv.config(), so every key the real .env may
+# hold must be blanked here or tests make live, billable calls.
+GEMINI_API_KEY=
+GEMINI_VI_TITLE_TRANSLATION_ENABLED=false
+GEMINI_VI_TITLE_MONTHLY_BUDGET_MICRO_USD=0
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@
 
 # Offline LLM diacritic experiment reservations and provider responses
 /tmp/llm-diacritic-ab/
+/tmp/llm-diacritic-ab-gemini/
 
 # Local implementation plans and working specifications
 /specs/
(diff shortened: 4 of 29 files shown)


Sinetheta:
## Follow-up: migration was not idempotent, and the spend guard had a hole

Two further fixes after testing against a real database.

### 1. The migration failed outright (`48e531a`)

```
ERROR: Could not migrate up 20260805000000_translation-unit-modes.js:
Index already exists with a different name: textHash_1_sourceLang_1_targetLang_1_mode_1
```

Mongoose's `autoIndex` builds the widened unique key from the schema, which declares no name, so Mongo generates one. The migration then demanded `translation_unit_mode_key` for the same keys — whichever ran second died. Every other index migration in this repo already sidesteps this by using Mongo's own generated name; this one invented a friendlier one.

Both names now match what Mongo generates, and indexes are located **by key pattern, not by name**, so the migration converges from any starting state: fresh DB, autoIndex-first, a legacy hand-picked name, or already applied. Same for `down()`.

This matters more than usual here: **staging and production share one database**, so every migration is run at least twice over the same data and the second run must be a no-op, not an error.

Verified against the database that was failing — `up`, `down`, `up` again, rows and the standard-mode backfill intact. Covered by a test that reproduces the exact autoIndex-first ordering.

### 2. An auth or rate-limit failure would have burned the whole allowance (`653fe38`)

Budget is reserved *before* each provider call, and the batch loop treated every error as a per-batch hiccup and continued. A wrong or revoked key, or a 429, would have marched through the entire census — reserving the per-call maximum each time, translating nothing, and quietly consuming the full monthly allowance before anyone read the log.

401/403/429 are now classified as unrecoverable: the pass stops at the failing batch, reports how many titles are unprocessed, and says it needs action rather than a retry. An exhausted allowance stops it the same way. Transient errors (500, malformed response) still skip just that batch. Re-running after a fix resumes rather than redoing.

### Verification

Backend suite **1069 passing / 0 failing** on a clean run (+2 tests).

One note on process: an intermediate run reported 40 failures. Those were 10s timeouts on DB-heavy specs caused by my own concurrent suite runs, not by these changes — re-run on a settled machine, all green. Flagging it because the number looks alarming in isolation.

PR #216, opened 2026-08-29
fix: exclude rawSource from getFeed's blueprint select

Sinetheta:
## Summary
- `getFeed` (`app/api/user-controller.ts:425`) was the sole list/feed query in `app/api` not excluding `rawSource` from its Mongo `select()` — every other list query uses `'-data -thumbnail -rawSource'` (documented convention in `spec/blueprintsv2-followups.md`'s BPv2 gotcha list).
- Fixed the select to match convention: `'-data -thumbnail -rawSource'`.

## Note on scope
`buildListItem` already constructs the response object with explicit fields (no spread), so `rawSource` never actually reached the `/api/feed` JSON body regardless of this select — verified empirically. The real cost was pulling up to ~2MB of verbatim blueprint text per feed row across the wire from Mongo unnecessarily, and the inconsistency with the other 6 call sites. The regression test spies on the underlying Mongoose query (`mongoose.Query.prototype.select`) rather than asserting on the response body, since the response shape can't actually expose this field — confirmed the test fails against the old select and passes against the new one.

## Test plan
- [x] New spec in `__tests__/api/social.test.ts` — verified it fails on the old select, passes on the new one
- [x] Full backend suite: 1101 passing
- [x] `npm run tsc` clean

### DIFF
--- a/__tests__/api/social.test.ts
+++ b/__tests__/api/social.test.ts
@@ -10,7 +10,8 @@ process.env.NODE_ENV = 'test';
 import { TestSetup } from '../setup/testSetup';
 import { BlueprintModel } from '../../app/api/models/blueprint';
 import { FollowModel } from '../../app/api/models/follow';
-import { Types } from 'mongoose';
+import mongoose, { Types } from 'mongoose';
+import sinon from 'sinon';
 
 describe('Profile, Follow, Feed API', function () {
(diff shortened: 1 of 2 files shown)


Sinetheta:
@coderabbit review

PR #223, opened 2026-09-05
fix: load more on Discover when a page does not fill the window

dytterud:
> [!NOTE]
> **AI-assisted change.** Written with the help of AI — Claude (Opus 5), via Claude Code. Please review it as you would any other change.

## The bug

Discover's infinite scroll was driven **solely** by a `window:scroll` handler, and `loadMore()` had exactly one caller. If the first page doesn't overflow the viewport there is nothing to scroll, no event fires, and the list is stranded on page one — no way to reach any further results.

The predicate itself was never wrong. With `scrollHeight ≈ innerHeight`, `scrolled > total - 300` evaluates to `innerHeight > innerHeight - 300` → true. The logic was one event away from doing the right thing; the event just never arrived.

## How reachable is it

Measured against production, not estimated. A page is `BROWSE_INCREMENT` (10) cards, and the grid is width-capped at 1132px / 3 columns — identical at 1920, 2560 and 3840 wide — so the document is **~1844px tall regardless of window width**. The threshold is therefore purely viewport height:

| Setup | CSS viewport height | Stuck? |
|---|---|---|
| 1920×1080 @100% | ~960 | No |
| 2560×1440 @100% | ~1320 | No |
| **3840×2160 @100%** | ~2040 | **Yes** |
| Portrait 1440×2560 | ~2440 | **Yes** |
| 2560×1440 @67% zoom | ~1970 | **Yes** |

Confirmed on the live site at 3840×2160: `scrollHeight` 2160, `innerHeight` 2160, `scrollable: false`. At 1843px there is exactly 1px of scroll, so the boundary is 1844.

Note many 4K users run 150% display scaling, which puts them at ~1440 CSS px and hides this — which is probably why it has gone unnoticed. Collapsing duplicates into fewer rows lowers the bar further.

## The fix

The scroll predicate is now shared (`maybeLoadMore`), and additionally:

- runs on `window:resize` — a window made taller can uncover the bottom without any scroll event
- is re-checked after every completed load, **deferred a task** so the just-appended cards are laid out before `scrollHeight` is measured; measuring synchronously reads the previous layout and loads one page too many

`MAX_AUTO_FILLS` bounds the loop so a page that somehow never grows cannot walk the whole corpus one request at a time. The existing `working` / `noMoreBlueprints` guards do the rest, and the counter resets in `resetPaging()` alongside the other paging state.

## Verification

- Three new specs. The key one — "loads another page when the content is shorter than the window" — was **confirmed to fail without the fix** (stashed the component, 1 failed / 105 passed), so it is a real regression test rather than a vacuous one. The other two are guards: don't over-load when content already overflows, and stop when the server reports no more results.
- Frontend suite: **1253 passed**, 2 skipped. `npm run lint` clean.
- `scrollHeight`/`innerHeight` are stubbed in the specs since jsdom does no real layout; the production measurements above are what ground the numbers.

**Not done:** the fix itself has not been exercised in a running browser — Discover needs the backend, and no database was running locally. The behaviour is covered by the failing-then-passing spec, but a manual check on a tall viewport would be worth doing before merge.

### DIFF
--- a/frontend/src/app/module-blueprint/components/browse-page/browse-page.component.spec.ts
+++ b/frontend/src/app/module-blueprint/components/browse-page/browse-page.component.spec.ts
@@ -170,6 +170,18 @@ describe("BrowsePageComponent", () => {
     // stubbed to describe the shape that matters: content shorter than the
     // window, which fires no scroll event and so used to strand the list on
     // page one.
+    // scrollHeight/innerHeight are shared globals — captured here and put back
+    // in afterEach, so a short viewport cannot leak into later tests in this
+    // file and schedule auto-fill work they never asked for.
+    const originalScrollHeight = Object.getOwnPropertyDescriptor(
+      Object.getPrototypeOf(document.documentElement),
+      "scrollHeight",
+    );
+    const originalInnerHeight = Object.getOwnPropertyDescriptor(
+      window,
+      "innerHeight",
+    );
+
     const setViewport = (docHeight: number, windowHeight: number) => {
       Object.defineProperty(document.documentElement, "scrollHeight", {
         value: docHeight,
@@ -190,6 +202,16 @@ describe("BrowsePageComponent", () => {
 
     afterEach(() => {
       vi.useRealTimers();
+      // Own-property overrides shadow the prototype getter; deleting restores it.
+      delete (document.documentElement as any).scrollHeight;
+      if (originalScrollHeight)
+        Object.defineProperty(
+          Object.getPrototypeOf(document.documentElement),
+          "scrollHeight",
+          originalScrollHeight,
+        );
+      if (originalInnerHeight)
+        Object.defineProperty(window, "innerHeight", originalInnerHeight);
     });
 
     it("loads another page when the content is shorter than the window", () => {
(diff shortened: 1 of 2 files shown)


dytterud:
Now verified in a browser, which the description previously listed as not done.

No local database was available (no Docker, no Mongo on this host), so the dev server's `/api` proxy was pointed at the production backend — anonymous read-only GETs — giving the local build real blueprints. That also makes for a clean A/B, since production runs the unfixed code.

**Identical viewport, 1920×2100:**

| | cards | scrollHeight | innerHeight | scrollable |
|---|---|---|---|---|
| Production (unfixed) | 10 | 2100 | 2100 | **false** — stranded |
| Local (this branch) | 20 | 3082 | 2100 | true |

**No change on ordinary screens.** At 1920×1000 the local build loads 10 cards, `scrollHeight` 1844, scrollable — it does not over-fetch when the first page already overflows, and it stopped at 20 rather than continuing to load once the content exceeded the window.

**Resize path** verified by dispatching a real `resize` event: 10 → 20 cards, `scrollHeight` 3082. Worth noting for anyone repeating this — changing the viewport via CDP emulation does *not* dispatch a `resize` event to the page, so emulation alone appears to show no effect; that is a harness artifact, not the handler.

The proxy change was local only and is not part of this branch.

Sinetheta:
Thank you @dytterud — this is a genuinely impressive bug report wrapped around a fix. The reachability analysis is the standout: measuring the real 1844px document height against production, tabulating exactly which viewports strand, and explaining why nobody noticed (4K at 150% display scaling hides it). The key spec was confirmed to fail without the fix, and the details show care — deferring the re-check one task so scrollHeight reflects the just-appended cards, coalescing resize bursts, cancelling on destroy, bounding the loop. The honest "Not done" note about not exercising it in a running browser is exactly the disclosure I want; I'll give it a tall-viewport check on the live site after deploy.

Merged as-is — thanks for both contributions.

PR #226, opened 2026-09-06
docs: keep the session-end PR in draft until manually tested

dytterud:
## What changed

`CLAUDE.md`'s session-end step now opens a **draft** PR rather than a ready one, and states
that it stays a draft until the feature is done — where done means manually tested by a
human, not "the tests pass".

Net +7 lines. Everything else in `## Work Session Lifecycle` is untouched: the opening
rationale, session start, and the committing rules.

```diff
-**Session end — push and open a PR (autonomously, without asking):**
+**Session end — push and open a _draft_ PR (autonomously, without asking):**
-3. `gh pr create` with a real description: …
+3. `gh pr create --draft` with a real description: … — leading with what has **not** been verified yet
-4. Report the PR URL as the session's final output
+4. Report the PR URL and the outstanding verification as the session's final output
+
+**The PR stays in draft until the feature is done, and done means manually tested by the
+user.** Passing tests are not done — they show the code does what its author intended, not
+that the feature works. Never take a PR out of draft; the user publishes it.
```

## Why

Claude opened #225 the moment `tsc`, 273 lib tests, 1291 frontend tests and lint went green, plus
a browser session It drove. The feature was not finished — it still needed a person to
read the code and exercise it by hand in the game these blueprints are for. #225 was
converted to draft, and the manual round trip later found the work sound, but only because
someone ran it.

The old instruction was not misread. It said to open a PR at session end and said nothing
about what "session end" requires, so passing checks became the finish line.

## Deliberately unchanged

Session end still pushes and opens a PR **autonomously** — the section exists because work
stranded on a local branch is invisible, and that reasoning still holds. The only change is
that the PR arrives as a draft, and publishing it is the user's signal that the feature is
done.

## Verification

Docs only — no code, no tests, no runtime behaviour. Opened as a normal PR at the user's
explicit request; under its own rule it would have gone up as a draft.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -910,12 +910,16 @@ Commit message format:
 
 Stage only relevant files — never `git add -A` blindly. Do not skip hooks (`--no-verify`).
 
-**Session end — push and open a PR (autonomously, without asking):**
+**Session end — push and open a _draft_ PR (autonomously, without asking):**
 
 1. Update any committed docs that describe shipped state (e.g. `agent/TODO.md`) so they reflect what this branch ships
 2. `git push -u origin <branch>`
-3. `gh pr create` with a real description: what shipped, design decisions and spec deviations, how it was verified (test counts, migrations run), and anything deferred
-4. Report the PR URL as the session's final output
+3. `gh pr create --draft` with a real description: what shipped, design decisions and spec deviations, how it was verified (test counts, migrations run), and anything deferred — leading with what has **not** been verified yet
+4. Report the PR URL and the outstanding verification as the session's final output
+
+**The PR stays in draft until the feature is done, and done means manually tested by the
+user.** Passing tests are not done — they show the code does what its author intended, not
+that the feature works. Never take a PR out of draft; the user publishes it.
 
 ## Important Instructions
 


Sinetheta:
Thanks for this — merging.

A bit of context on why it was the other way round. When I set that section up I was the only person working in the repo, and I'd tuned it for my own throughput: let the agents open PRs the moment checks were green, let CI and CodeRabbit churn on everything, and I'd sort out what was actually finished afterwards. That optimised my time at the cost of everyone else's, which was a fine trade while "everyone else" was nobody.

That's not the situation any more, and this is the better default for the repo's long-term health. Draft until a person has actually exercised the feature is the right bar, and the repo was already half set up for it — the Slack alert skips drafts and fires on ready-for-review, so nothing else needed changing.

One thing worth knowing: CodeRabbit skips draft PRs by default, so the bot review will now land when you mark a PR ready rather than when you open it. I think that's fine, and probably better — review after manual testing rather than before.

More generally: any change that makes it easier or saner to contribute here is a good change, and I'd rather have the friction on my side than on a contributor's. So thanks, and keep them coming.

PR #227, opened 2026-09-06
chore: force LF line endings via .gitattributes

dytterud:
## What shipped

Adds `* text=auto eol=lf` to `.gitattributes`.

## Why

`.prettierrc.json` sets `"endOfLine": "lf"`, but nothing enforces LF on checkout. A Windows
checkout with `core.autocrlf=true` (the common default) converts every file to CRLF on
checkout, which fights that config directly. Result: `npm run format:check` flags nearly the
entire repo (274 files) as needing reformatting — not a real diff, just line-ending noise that
buries any actual formatting problems under it.

Confirmed while working on `feat/threshold-sensor-settings`: `format:check` flagged 274 files;
after normalizing line endings for comparison, only 2 had real content differences (both fixed
separately, unrelated to this change).

## Scope

`.gitattributes` only governs *future* checkouts and clones — it does not retroactively fix
line endings already on disk or in history. Renormalizing the existing working tree
(`git add --renormalize .`) is a separate, much larger, deliberate change (touches all 274
files at once) and is intentionally not part of this PR.

## Verification

Docs/config only, no code. Read the diff; no tests to run.


Sinetheta:
Thanks — merged. Good catch, and the right kind of fix: one line that stops a whole class of noise for every future Windows contributor.

Small correction to the numbers for the record: the 274 files were your checkout after autocrlf conversion, not the repository. The index was already ~98% LF, with just 21 files actually stored as CRLF. So the deferred renormalize turned out to be tiny, and it's up as [#228](https://github.com/blueprintnotincluded/blueprintnotincluded/pull/228) to land right behind this one so those files convert once, in a commit that says so, rather than one at a time inside unrelated PRs.

More broadly: this repo carries a lot of history. It was built and then generously handed over with everything intact, and that's the only reason any of us have it to work on — but it also means there are plenty of things like this that were never standardised because they never needed to be for one person on one machine. None of it is sacred. If you spot something fundamental that's worth fixing, tooling, structure, conventions, whatever, I'd rather see the PR than have you assume it's off limits.

PR #231, opened 2026-09-09
feat(editor): edit Critter Sensor settings (buildingData)

dytterud:
## What shipped

Extends the threshold-sensor settings editing (PR #225) to the **Critter Sensor** (`LogicCritterCountSensor`).

A copied in-game Critter Sensor carries three `buildingData` keys:
- `Switch` — sampled output at copy time (stowaway, same as threshold sensors)
- `LogicCritterCountSensor` — the real settings: `countThreshold`, `activateOnGreaterThan`, `countCritters`, `countEggs`
- `IThresholdSwitch` — a **pure redundant echo**: in the decompiled base game `LogicCritterCountSensor.Threshold` is `get => countThreshold; set => countThreshold = (int)value`, and `ActivateAboveThreshold` aliases `activateOnGreaterThan`

Before this change the panel rendered all three — an editable "On" checkbox for the stowaway `Switch`, and **duplicate** Threshold/direction rows (one pair from each key), with the `IThresholdSwitch.Threshold` shown as a raw unitless float.

Now:
- `resolveSettingDescriptors` suppresses `Switch` and `IThresholdSwitch` to `[]` for this prefab (both round-trip untouched). Its own key is the sole editable group: `countThreshold` (0–64, `RangeMax` from the decompile), above/below segmented control, count critters, count eggs.
- `LogicCritterCountSensor` is now creatable from scratch (`CREATABLE_SETTINGS`) with the real Klei defaults (`countThreshold` 0, the three bools true — from the decompiled `[Serialize]` field initializers), so an editor-placed sensor gets a "Critter count — (Not set)" row + **Set**.
- The threshold "not set"/Clear machinery is generalized to `primarySettingsKey(prefabId)` — threshold sensor → `IThresholdSwitch` (labelled by the measured quantity), Critter Sensor → its own key. Component getters `thresholdLabel`/`canClearThreshold`/`setThreshold`/`clearThreshold` → `primaryUnsetLabel`/`canClearPrimary`/`setPrimary`/`clearPrimary`.
- **Echo consistency**: an edit to `countThreshold`/`activateOnGreaterThan` is mirrored onto an existing `IThresholdSwitch` echo (`redundantEchoField`, applied in `BlueprintItem.setBuildingSetting` — the single write chokepoint). Without this, a stale echo could clobber the fresh value on the mod's key-apply pass (`IThresholdSwitch.TryApplyData` sets `countThreshold = (int)Threshold`). **Clear** drops both `LogicCritterCountSensor` and the echo (keeps `Switch`, matching the threshold-sensor precedent).

## Design notes / research

Verified against three sources, all agreeing:
- mod source `BlueprintsV2/BlueprintData/DataTransferHelpers.cs` (`DataTransfer_LogicCritterCountSensor`, `DataTransfer_IThresholdSwitch`)
- decompiled base game `LogicCritterCountSensor` (ilspycmd): `RangeMin 0` / `RangeMax 64`, `[Serialize]` defaults `countEggs=countCritters=activateOnGreaterThan=true`, `countThreshold` int (0)
- a real in-game export: keys, casing, and value types byte-for-byte as expected (`countThreshold` int, `Threshold` float echo)

`LogicCritterCountSensor` stays out of `THRESHOLD_SENSORS` deliberately — it needs no unit conversion and its `IThresholdSwitch` is suppressed rather than rewritten.

## Verification

- `npm run test` — lib/backend suite; **10 new tests** (`__tests__/lib/building-settings.test.ts`): suppression, catalogue bounds, creatable defaults, `primarySettingsKey`, and echo-mirror round-trips. The 4 unrelated failures on this machine (WorkOS API timeouts, Windows path-separator asset assertions) are pre-existing and present on master.
- `npm run tsc` — clean.
- `cd frontend && npm test -- --include='**/building-settings.component.spec.ts'` — **37 passing** (4 new/updated critter specs: one clean group / no stowaway / no duplicate, echo mirror through the input, set + clear both keys; plus mechanical getter renames in the threshold specs).
- `cd frontend && npm run lint` + `tsc -p tsconfig.app.json` — clean.
- Not done: live click-through in the editor (the running dev server on this machine is another checkout's container). The component specs assert the rendered DOM (row list, checkbox count, button text/presence, clear behaviour).

## Docs

Updated `CLAUDE.md` "Threshold sensors" section (critter-sensor bullet rewritten, "not set"/`primarySettingsKey`, creatable-from-scratch list) and the status block.


dytterud:
### Status

- **Manually tested in-game ✅** — loaded a blueprint with a Critter Sensor (threshold 5, "below", count critters on, count eggs off) through the BlueprintsV2 mod; all values came through correctly on the placed building. Also verified live in the editor: four editable rows, no stray "On" row, no duplicate threshold; editing the threshold mirrors onto the `IThresholdSwitch` echo; Clear drops both keys and keeps `Switch`; Set restores Klei's defaults.
- **Expected `CLAUDE.md` conflict with #230** (`docs: split root CLAUDE.md into nested per-directory docs`). Code changes don't conflict — #230 is docs-only. This PR edits the "Threshold sensors" section + status block in root `CLAUDE.md`; #230 relocates that section to `lib/src/blueprint/CLAUDE.md`. If #230 merges first, rebase and move the three doc edits (critter-sensor bullet, `primarySettingsKey` in the "not set" bullet, `LogicCritterCountSensor` in the creatable list) into `lib/src/blueprint/CLAUDE.md`, dropping the status-block edit if that block is gone.

Sinetheta:
Thanks @dytterud — this is follow-up #1 from #225 and you went further than I asked. I suggested "hiding or mirroring the duplicate"; you did both, and the reasoning for why both are needed is the best part of the PR.

What I want to call out:

- The echo mirror never creates the echo. redundantEchoField only updates an IThresholdSwitch entry the file already carries, so a sensor placed in the editor exports one clean key and an in-game copy stays internally consistent. That's the conservative choice and it's the correct one — synthesizing the alias would have been the easy version and it would have been wrong.
- Suppression goes through resolveSettingDescriptors, not a special case in the component. Because otherKeys gates on isKnownSettingsKey rather than descriptor count, the hidden echo and the stowaway Switch stay invisible and uncounted, with no new plumbing. You found the seam that already existed instead of cutting a new one.
- Generalizing to primarySettingsKey rather than branching on the critter sensor in the component. The panel now has one concept of "the canonical key this building pins or doesn't", and the threshold sensors fall out of it unchanged. That's the version of this change that ages well.
- The tests are load-bearing. I stubbed out the echo mirror in setBuildingSetting and two of your lib tests fail immediately with the stale value. Verifying the shape against a real in-game export — keys, casing, value types — is the check CI can't do, and it's the one that mattered.

On the judgment call: Clear on a critter sensor now also discards countCritters/countEggs. That's precisely why I deferred this rather than solving it badly, and I'm accepting your answer — the fields genuinely can't be separated, and renaming the button with a re-worded tooltip is honest about what it does rather than pretending otherwise.

Non-blocking follow-ups, if you want them:

1. Clear ${label} — "Clear Pressure" / "Clear Critter count" — would keep the specificity the old "Clear threshold" had, for about one line.
2. max: 64 and the four creatable defaults rest on the decompile rather than the game. When you're next in a sandbox, confirming them the way you confirmed the sensors in #225 would close the loop. Low stakes either way — the range is soft and the defaults only touch a from-scratch sensor.
3. Still open from #225: LogicWattageSensor, an issue for the two unit discrepancies, and element sensors.

Thanks again — keep them coming.

PR #232, opened 2026-09-10
proposal(editor): stop the build menu switching overlay when the item is already visible

dytterud:
**Draft / proposal — this is a product question, not a bug fix. Needs a maintainer call before it goes anywhere.**

## What happened to the original framing

This PR opened claiming to fix a reported symptom: *"pressing B on an automation wire in the Automation overlay makes all other items under the mouse included in that overlay."*

@Sinetheta was right that the code doesn't do that, and the reason is structural: `isOverlaySecondary` is `overlay == this.overlay`, so it can only ever hold when the item's own overlay *already equals* the current one. It can never be what stops a switch **away** from Automation. The only switch this gate actually prevents is `isOverlayPrimary(current)` for a Building-layer item whose viewMode differs from `current` — i.e. Base, or Room collapsing to Base.

**The reported bug is separately fixed.** `aa0bb8d1` (element brush / B-key sampling) rewrote `changeToBuildToolFromSelection` so B samples the front-most item under the cursor. Confirmed by hand on production — which does *not* have this PR — and the symptom is gone there. So nothing in this PR is needed for that issue.

## What this actually changes

In the Buildings overlay, picking a Gas Pump / Battery / LogicSwitch from the build menu (or copying one with B) currently pulls the view into Gas / Power / Automation. 204 buildings in the shipped database are Building-layer with a non-Base viewMode overlay (75 Power, 52 Liquid, 33 Automation, 31 Gas, 13 Conveyor). This gate makes the view stay put whenever the picked item already renders opaque where you are.

Note `changeItem` is the build menu's path (`uiItemChanged`), so this is not limited to copy/select as the original title said.

## Why this needs a decision rather than a merge

`setOverlayForItem` dates to the **initial commit**. The unconditional switch is original, deliberate design — not an oversight. And there's a real argument it should stay: when you pick a Gas Pump you plausibly want the Ventilation overlay, because what matters for placing it is seeing the *pipe network*, not seeing the pump. "The item is already visible" is the wrong test if the point was never the item.

I don't have evidence either way about what the game does here, and I shouldn't have asserted parity with it earlier without checking.

So: **is the build menu's overlay-follow still wanted?**
- If yes — close this.
- If it's wanted only when it reveals something you couldn't already see — this is that, and it needs a description, not a rewrite.
- If the Room case alone is wrong (Room silently collapsing to Base) — that can be narrowed to a much smaller change.

## Contents

- `lib/src/drawing/camera-service.ts` — the gate, plus a corrected comment.
- `__tests__/lib/camera-service.test.ts` — my original specs, plus @Sinetheta's `34496853`, which added the two that actually fail without the gate and a contract spec pinning the hand-written fakes to real database buildings. Kept as-is; it's the better test of the two.
- Current `master` merged in; `npm run build:lib` clean.

## Still open from the same report

The other two items in the source issue are untouched: automation buildings rendering behind automation wire (a depth-band bug in `BlueprintItem.cameraChanged`), and missing port markers on logic gates.

### DIFF
--- a/__tests__/lib/camera-service.test.ts
+++ b/__tests__/lib/camera-service.test.ts
@@ -1,6 +1,6 @@
 import { describe, it } from 'mocha';
 import { expect } from 'chai';
-import { CameraService, Vector2 } from '../../lib/index';
+import { CameraService, Overlay, Vector2 } from '../../lib/index';
 
 // Pure camera math — no PIXI needed: the constructor only stores the
 // container reference, and pinchZoom/updateZoom/changeZoom never touch it.
@@ -84,5 +84,70 @@ describe('CameraService.pinchZoom', function () {
     camera.zoom(1, new Vector2(0, 0));
     for (let frame = 0; frame < 300; frame++) camera.updateZoom();
     expect(camera.currentZoom).to.be.closeTo(64, 1e-6);
+  });
+});
+
+describe('CameraService.setOverlayForItem', function () {
+  // Only the three members setOverlayForItem touches — a real OniItem would drag
+  // in the whole element/sprite/database bootstrap.
+  const fakeItem = (
+    overlay: Overlay,
+    opaqueIn: Overlay[]
+  ): any => ({
+    overlay,
+    isOverlayPrimary: (o: Overlay) => opaqueIn.includes(o),
+    isOverlaySecondary: (o: Overlay) => o === overlay,
+  });
+
+  // Building-layer device: opaque in its viewMode overlay (secondary) and in Base
+  // (primary), grey everywhere else — like LogicSwitch / a Logic*Sensor.
+  const automationDevice = fakeItem(Overlay.Automation, [Overlay.Base]);
+  // Wire: primary in its own overlay only — like LogicWire.
+  const automationWire = fakeItem(Overlay.Automation, [Overlay.Automation]);
+  // Plain building: Base only — like a Bed or a tile.
+  const plainBuilding = fakeItem(Overlay.Base, [Overlay.Base]);
+  // Gas building: opaque in Gas only.
+  const gasBuilding = fakeItem(Overlay.Gas, [Overlay.Gas]);
+
+  it('switches from the initial None overlay to the item overlay', function () {
+    const camera = makeCamera();
+    expect(camera.overlay).to.equal(Overlay.None);
+    camera.setOverlayForItem(automationDevice);
+    expect(camera.overlay).to.equal(Overlay.Automation);
+  });
+
+  it('does not leave the Automation overlay when copying an automation wire', function () {
+    const camera = makeCamera();
+    camera.overlay = Overlay.Automation;
+    camera.setOverlayForItem(automationWire);
+    expect(camera.overlay).to.equal(Overlay.Automation);
+  });
+
+  it('does not leave the Automation overlay when copying an automation device (secondary)', function () {
+    const camera = makeCamera();
+    camera.overlay = Overlay.Automation;
+    camera.setOverlayForItem(automationDevice);
+    expect(camera.overlay).to.equal(Overlay.Automation);
+  });
+
+  it('leaves the Automation overlay when copying a plain building it would grey out', function () {
+    const camera = makeCamera();
+    camera.overlay = Overlay.Automation;
+    camera.setOverlayForItem(plainBuilding);
+    expect(camera.overlay).to.equal(Overlay.Base);
+  });
+
+  it('switches to the item overlay when the current one would grey it out', function () {
+    const camera = makeCamera();
+    camera.overlay = Overlay.Power;
+    camera.setOverlayForItem(gasBuilding);
+    expect(camera.overlay).to.equal(Overlay.Gas);
+  });
+
+  it('treats the Room overlay as Base for the compatibility check', function () {
+    const camera = makeCamera();
+    camera.overlay = Overlay.Room;
+    camera.setOverlayForItem(plainBuilding); // opaque in Base -> Room counts as Base
+    expect(camera.overlay).to.equal(Overlay.Room);
   });
 });
--- a/lib/src/drawing/camera-service.ts
+++ b/lib/src/drawing/camera-service.ts
@@ -126,6 +126,14 @@ export class CameraService {
   }
 
   setOverlayForItem(item: OniItem) {
+    // Copy-building ("B") and selection call this to reveal the picked item. If
+    // the overlay we're already in renders it opaque — its own viewMode overlay,
+    // or Base for a plain building — stay put; only switch when the current
+    // overlay would grey it out. Without this, copying/selecting a plain tile
+    // while in the Automation overlay flips the whole view to Base.
+    // Mirror BlueprintItem.cameraChanged's Room->Base collapse.
+    const current = this.overlay_ === Overlay.Room ? Overlay.Base : this.overlay_;
+    if (item.isOverlayPrimary(current) || item.isOverlaySecondary(current)) return;
     this.overlay = item.overlay;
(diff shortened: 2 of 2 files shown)


Sinetheta:
@dytterud — apologies, I got ahead of you on this one. I took it out of draft and started reviewing before you'd said it was ready, and then pushed a commit to your branch. I also rebased and force-pushed #231 earlier. I was trying to keep your merges moving cleanly, but pushing to someone else's PR isn't mine to do by default and I should have asked. Sorry.

Part of why I'm backing off: I think the draft flag was doing real work, because there's a mismatch between what this says and what it does.

The description's headline case is "copying a plain building while in the Automation overlay killed the automation highlight." That case isn't changed here. A plain building genuinely renders at 0.3 alpha in Automation, so the new rule still switches you to Base — and your own spec, _leaves the Automation overlay when copying a plain building it would grey out_, asserts exactly that.

What the change actually does is the other direction: it stops the view leaving **Base** and **Room**. I measured it against the database — 204 buildings are Building-layer with a non-Base viewMode overlay (75 Power, 52 Liquid, 33 Automation, 31 Gas, 13 Conveyor), and for all of them, picking one while in Base used to yank the view into their overlay and now doesn't. Your manual-verification note describes that correctly ("from the Buildings overlay… pressing B no longer flips to Automation"); it's the root-cause paragraph above it that doesn't match.

Related: `changeItem` is also the build menu's path (`uiItemChanged`), so this changes overlay behaviour when picking from the menu too, not only on copy/select as the title says.

So — which one did you mean? If Base/Room preservation is the intent, this is good and the title and description need to catch up. If the Automation-overlay hijack was the target, this doesn't get there yet, and I'd want to understand the Refinaria repro better before guessing at a rule.

The commit I pushed (`34496853`) is tests only, no source changes. The reason I wrote it: five of your six new specs pass against unfixed master — they assert cases where the old unconditional `this.overlay = item.overlay` agrees with the new gate — so only the Room one was pinning the change. It adds the two that fail without your fix, plus a contract spec tying the hand-written fakes to real buildings. Drop it, squash it, or rewrite it as you see fit; it's your branch and I'll keep my hands off it from here.

dytterud:
Thanks — you were right, and the correction went further than fixing the description.

Confirmed your read independently: `isOverlaySecondary` is `overlay == this.overlay`, so it only ever holds when the item's own overlay already equals the current one. It structurally *cannot* be what stops a switch away from Automation. The only switch the gate prevents is `isOverlayPrimary(current)` for a Building-layer item — Base, or Room collapsing to Base. Exactly the direction you measured.

Two things followed from that.

**The reported bug was already fixed elsewhere.** `aa0bb8d1` (your B-key sampling work) made `changeToBuildToolFromSelection` sample the front-most item under the cursor. I had the reporter check the Refinaria repro on production — which doesn't have this PR — and it's gone there. So this PR was never load-bearing for the issue it was filed against.

**And then the part I got wrong on my own.** Digging into whether the remaining behaviour change is worth having, `setOverlayForItem` turns out to date to the initial commit. The unconditional switch is deliberate original design, and there's a decent argument it should stay: picking a Gas Pump plausibly *should* drop you into Ventilation, because what you need for placing it is the pipe network, not the pump. My "the item is already visible, so switching has no benefit" reasoning tests the wrong thing. I'd also written "matches ONI copy-building" in my notes without ever verifying it — that was a guess I presented as fact.

So I've retitled this as a proposal and put the question in the body rather than pretending it's a fix. It's genuinely your call whether the build menu's overlay-follow is still wanted; if it is, close this and I'll have no argument.

No hard feelings at all about the push or the draft flag — the commit you added is the better test (mine had five specs that passed against unfixed master, which is exactly the problem you named). I've kept `34496853` as-is.

Still outstanding from the original report, untouched here: automation buildings rendering behind automation wire (a depth-band issue in `BlueprintItem.cameraChanged` — Building-layer devices land at `zIndex + 50` = 69 while the wire sits at `zIndex + 100` = 113), and missing port markers on logic gates.

Sinetheta:
Thanks for the rewrite — and for digging out the initial-commit history and the unverified parity note yourself. Retitling this as a proposal rather than defending the original framing is the right instinct. Your read of `isOverlaySecondary` is correct, and with your commit plus mine there are now three specs that fail without the gate rather than the one.

I went and got the answer to the question you asked, against https://github.com/Kupie/ONI_Decomp at `ChangeList = 744825` — the same build as our own database baseline, `U59-744825-SCRPAN`. `InterfaceTool.OnActivateTool`:

```
if (OverlayScreen.Instance != null && this.viewMode != OverlayModes.None.ID && OverlayScreen.Instance.mode != this.viewMode)
{
    OverlayScreen.Instance.ToggleOverlay(this.viewMode, true);
}
```

`BuildTool.Activate` sets `this.viewMode = def.ViewMode`, and Copy Building routes through `PlanScreen.CopyBuildingOrder` → `OnSelectBuilding` → that same call, so the game shares one path there too, exactly like `changeItem`. It switches unconditionally, with no test for whether you could already see the thing — and identically in mirrors from 2022, so it isn't version drift. The overlay-follow is still wanted, and this gate moves us away from the game for precisely those 204 buildings. On that question alone I'd close this.

Except the same code answers a question neither of us asked. `BuildingDef.ViewMode` defaults to `OverlayModes.None.ID` (Bed and Ladder set nothing; Gas Pump sets GasConduits), and our own `overlayFromViewMode` maps a null viewMode to `Overlay.Base`. So `item.overlay === Overlay.Base` *is* ONI's `ViewMode == None`, and for those the game doesn't switch at all. Picking a Bed while you're in Automation leaves you in Automation. We drop you to Base. That's the original report's actual complaint, and neither master nor this PR fixes it.

So the change I'd take is the opposite gate: `if (item.overlay === Overlay.Base) return;`. One catch — `build-tool.component`'s ngOnInit loads a Tile, and that call is what moves the camera off `Overlay.None`, so a bare early return boots the editor with no overlay. The contract spec on this branch already pins that, conveniently.

It's your branch and your find. Want to take that, or shall I?

dytterud:
Taking it — #240.

You're right on both halves, and the second one is the find. `InterfaceTool.OnActivateTool` has no visibility test, and Copy Building shares the path, so the overlay-follow is deliberate and this PR's gate was moving us away from the game for 243 buildings. Nothing of it survives in #240.

The `ViewMode == None` gate is in, with your `Overlay.None` catch. Worth stating why that catch is load-bearing beyond the boot sequence: `Overlay.None` isn't neutral — nothing is primary or secondary there, so `cameraChanged` renders the whole blueprint at 0.3 alpha on bare `zIndex`, which also changes select-tool frontmost resolution since it picks max `depth`. A bare early return doesn't just skip an overlay, it boots into that.

Two divergences from the decompile I couldn't avoid, both in #240's body:

`item.overlay === Overlay.Base` is wider than ONI's test. `getRealOverlay` collapses the overlays we don't render — Oxygen, Decor, Light, Temperature, Rooms, Radiation — into Base as well, so 64 further buildings stop switching where the game switches. We have nowhere to switch them to; today they yank you to Base. Better, but not parity.

And the element brush reaches `setOverlayForItem` through the same `changeItem`. The synthesized `Element` OniItem carries `Overlay.Base`, so painting an element no longer pulls you to Base — it paints into whatever overlay you were in, where elements sit at 0.3 alpha. Same behaviour class as placing a Bed in Automation, which the game accepts, so I left it alone and pinned it in a spec rather than inventing an `isElement` exemption you didn't ask for. If you want it exempted it's one line.

Your `34496853` is on the commit. The assertions had to invert, but the contract-spec idea is the reason that block tests the code instead of a model of it, and I kept it — it's what pins that no `OniItem` can carry `Overlay.None`.

Closing this in favour of #240.

PR #234, opened 2026-09-12
Blueprint previews: draw terrain, bake a real cell grid

Sinetheta:
## Summary

Two independent changes to the server preview pipeline, from user feedback: geysers were missing from blueprint card/hero previews, and the grid behind them was a fixed 32px CSS pattern unrelated to the blueprint's actual scale (misleading — a 2x3 building could read as 4x6).

**B1 — terrain in server previews**
- Lifted the pure terrain footprint/icon-placement geometry out of the frontend's `draw-terrain-overlay.ts` into `lib/src/drawing/terrain-markers.ts` (the same move `note-markers.ts` made) so the server preview worker — which cannot import from `frontend/` — can draw the same thing. Selection-only concerns (white selected outline, magenta active-tile highlight, sprite pooling) stayed in the frontend overlay.
- `preview-render-worker.ts` now draws terrain features with their own icon texture cache, z-ordered above buildings and below world notes (matching the client's existing export order, which I also reordered — it had been drawing terrain *over* notes).
- Added a geyser to the worker's `--smoke` fixture so the deploy-image smoke run exercises the path.

**B2 — real cell-pitch grid**
- The worker now returns pixel framing (`tileSize`, `offsetPx`) alongside the rendered master.
- Variant derivation composites a phase-aligned grid line SVG into `card.webp`/`hero.webp` only, scaled per variant — `og.png` is untouched, since its transparent-trim step depends on the master staying fully transparent.
- The grid is composited *behind* the (still-transparent) artwork rather than baked into an opaque background, so each skin's `--bni-thumbbg`/ground colour still shows through — only the old decorative CSS grid is gone from the loaded-card/hero states. It's kept for the skeleton/no-thumbnail placeholder states, which have no real render to derive a grid from.
- Framing is also stored on the Mongo `previewimages` rows (nullable `renderVersion`/framing-bearing rows); missing framing (old rows, or a legacy `renderMasterFn` test double returning a plain `Buffer`) means no grid, never an error.
- Added a `PREVIEW_RENDER_VERSION` gate so existing durable rows are treated as stale exactly once after this deploy and regenerate with the grid, rather than keep serving pre-change bytes indefinitely under the ordinary `modifiedAt` freshness rule. Scoped to the L2 Mongo rows only — L1 disk cache is ephemeral (wiped every redeploy per its own file header) so it didn't need a version marker, and versioning the file path would have broken every existing disk-cache-path assertion in `preview.test.ts` for no benefit.

## Design decisions / deviations from the plan

- Framing is *not* persisted as a reusable "stored master" — there's no code path that re-derives variants from a previously-rendered master without a fresh worker render, so the Mongo fields are informational/forward-compatible rather than load-bearing at serve time (the grid is already baked into the served bytes).
- Skipped a framing sidecar on the L1 disk cache for the same reason — nothing reads it back.

## Out of scope (mentioned per the plan)

- Planning Tool shapes are still missing from the server preview (same gap terrain had, tracked as a separate follow-up — the CLAUDE.md "Known gap" bullet now calls this out explicitly).
- The dead `ExportImageOptions.gridLines` checkbox in the export-images dialog (`blueprint-service.ts`) — unrelated to this pipeline, reads nowhere, left alone.

## Verification

- ✅ `npm run tsc` clean
- ✅ `npm run test` — 1120 passing (added `__tests__/lib/terrain-markers.test.ts` for the lifted geometry, and a `cell-pitch grid (framing)` describe block in `__tests__/api/preview.test.ts` covering: grid baked into card/hero when framing present, no grid when framing absent, og.png byte-identical with/without framing while card differs, and a render-version staleness test)
- ✅ `npm run build` clean (worker ships compiled; verified the compiled artifact too)
- ✅ `cd frontend && npm test` — 1284 passing (moved the pure-geometry specs out of `draw-terrain-overlay.spec.ts`, kept the `DrawTerrainOverlay` pooling/selection specs)
- ✅ `cd frontend && npm run lint` clean
- ✅ Worker `--smoke` passes with the new geyser fixture (19% pixel coverage)
- ✅ Rendered a real blueprint (buildings + a geyser + a world note) through the actual forked worker and inspected the output directly: geyser icon and dashed footprint render correctly, world note sits above it, the composited grid is visibly aligned to the real cell pitch (a 1x1 Tile building spans exactly one grid square), and `og.png` stays a plain white letterbox with no grid

### Not yet verified (needs a real browser session against a running app)

- [ ] Open a blueprint with a geyser, save it, open its details page: hero shows the geyser icon in its dashed footprint below any notes; the Discover card shows it too
- [ ] On that card, count grid squares across a known 2x3 building: three
- [ ] Share-link preview (`og.png`) unchanged: white letterbox, no grid
- [ ] Switch skins: the ground colour behind the preview follows the skin
- [ ] A pre-existing (pre-deploy) blueprint's card actually regenerates with the grid on first view after deploy, rather than needing a manual cache-bust

This PR stays in draft until the above is manually tested.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


Sinetheta:
@coderabbitai review

Sinetheta:
@coderabbitai review

PR #235, opened 2026-09-12
Element annotations UX: editable brush, B-key sampling, overlay-aware fading

Sinetheta:
## Not yet verified

I could not do interactive manual testing in this environment (no browser). Everything below is verified by automated tests only — **please run the manual test script** at the bottom before taking this out of draft.

## What shipped

Three coupled fixes for element annotations, from user feedback (full plan: `specs/element-annotations-ux-plan.md`, gitignored — not in this diff).

### 1. The element brush stays editable after placement

- `NotesTool.mouseDown` no longer selects the note it just placed (and explicitly clears any prior selection), so the edit panel keeps showing the *pending* template. Editing the element/mass/temp fields now configures the next click instead of silently editing the note that was just placed.
- `build-tool.component.ts`'s `changeElement` now calls `BlueprintItem.setElement(...)` for real, instead of only flagging a camera reload (the actual `setElement` call was dead/commented-out code).
- **The "B resets the brush to Vacuum" bug**: pressing "B" re-enters the build tool, which unconditionally fired `toolChanged(build)` → rebuilt the template from scratch even when nothing changed. `ComponentSideBuildToolComponent.toolChanged`/`ensureBuildItem` now only rebuilds when the incoming item's id actually differs from what's already loaded, so re-entering the tool (via B or otherwise) keeps whatever element/temperature/material was already configured.
- **The "can't repaint over a cell" bug**: `BuildTool.updateBuildCandidateResult` was blocking a build whenever anything shared the same `objectLayer`, and the synthetic `Element` OniItem's default layer (0) collided with itself. Decision: rather than giving `Element` its own `objectLayer` sentinel (risks colliding with a real game-derived value on some other building), I special-cased element items to skip that collision check entirely — they're annotations, not construction, so they should never be blocked by what's already on the tile. `BuildTool.build()` now removes any existing element cell(s) at the target tile before adding the new one, so painting replaces rather than stacks.

### 2. "B" samples what's under the cursor when nothing is selected

`ToolService.changeToBuildToolFromSelection` (still reachable via the `toolBuild` shortcut, relabeled "Build tool / sample under cursor") now:
- keeps today's behavior when something is select-tool-selected (clone it into the build tool);
- otherwise checks a selected world note (element notes only — text-note sampling is optional per the plan and deferred, see below);
- otherwise samples the hovered tile: the frontmost real blueprint item (building or element cell, by depth — same precedence `SelectTool` already uses) clones into the build tool; an element world note switches to the notes tool with that note's id/mass/temp as the pending brush;
- otherwise falls through to today's "just switch to build tool, no copy" behavior.

**Deferred (optional per the plan):** sampling a *text* note's symbol/color into the notes tool. Left out to keep this PR's surface area focused; it's a small addition on top of the same machinery if wanted later.

### 3. Overlay-aware alpha/depth for element cells and element notes, plus a "Show notes" toggle

- `BlueprintItemElement.cameraChanged` previously hard-coded alpha to `1` always, and depth had a gas-only `+50` hack. It's now one rule for every state (gas/liquid/vacuum share the +100 primary / +50 own-overlay / raw-elsewhere pattern conduits use; solid always sits at `Backwall` depth), with alpha `1` in the primary set and `0.3` (the same constant buildings use, now exported as `OVERLAY_DIMMED_ALPHA`) elsewhere — instead of hiding the sprite outright. Room now maps to Base (matching how buildings treat it), and `Display.None` renders like `Display.solid` (matching `DrawPart.prepareVisibilityBasedOnDisplay`), fixing two more "cells vanish" cases.
- The overlay-membership rule (`elementOwnOverlay`/`isElementOverlayPrimary`, in `lib/src/b-export/b-element.ts` to avoid a `drawing/` ↔ `blueprint/` import cycle) is shared between element cells and element world notes, so a `noteOverlayAlpha` helper fades an element note's marker exactly like the matching cell would. This is computed **per frame**, not cached alongside the marker/color (which are cached by note-array identity) — `saveImages()` draws one shared `DrawNotesOverlay` instance across several exported overlays with the *same* note array, so a cached alpha would only ever reflect the first exported overlay.
- Added `WorldNoteService.visible` + a "Show notes" checkbox in the note edit panel, mirroring `TerrainAnnotationService.visible`/its checkbox exactly: view-only, hides the pin layer on the live canvas only (`component-canvas.component.ts`, gated on `!forceSize`), never reaches exports/thumbnails/the server preview worker, and toggling off clears any active selection (an invisible thing can't stay selected/editable).

## Design decisions

- Kept the Notes-tool "select an existing note on click" behavior for every other tool (Select, Build, ...) unchanged — the top-of-stack note hit-test in `component-canvas.mouseClick` is now skipped only while the Notes tool itself is active, since its own `mouseDown` already resolves that exact click.
- Didn't rename `ToolService.changeToBuildToolFromSelection` (only its behavior/doc-comment/shortcut label) — it's internal and only referenced by its own spec.
- Didn't touch `paintElement()` (the "Paint Element" build-menu button) — picking a fresh element type from the menu resetting to a Water default is consistent with how picking any other building type from the menu resets its brush; the "B" bug was specifically about *re-entering* the tool doing that unconditionally, not about deliberately picking a new item.

## Verified (automated)

- `npm run tsc` — clean
- Backend: `npm run test` — 1115 passing (incl. new `__tests__/lib/blueprint-item-element.test.ts`)
- Frontend: `cd frontend && npm test` — 1322 passing, 2 skipped (pre-existing)
- `cd frontend && npm run lint` — clean

New/updated specs: `notes-tool.spec.ts`, `note-edit-panel.component.spec.ts` (new file), `tool-service.spec.ts`, `build-tool.spec.ts`, `build-tool.component.spec.ts`, `world-note.service.spec.ts`, `draw-notes-overlay.spec.ts` (incl. `noteOverlayAlpha` and a `DrawNotesOverlay` class-level regression test for the per-frame-not-cached alpha), and the new backend `__tests__/lib/blueprint-item-element.test.ts`.

## Manual test script

1. Notes tool → Element → pick Chlorine, 500 g. Click three cells. All three are Chlorine 500 g. Change the picker to Oxygen. Click a fourth cell. It is Oxygen; the first three are unchanged.
2. Hover a placed Chlorine note with nothing selected, press B. The panel shows Chlorine 500 g as the pending note; clicking places another.
3. Hover a building, press B: today's copy behaviour. Select a building, hover elsewhere, press B: the selection wins.
4. Paint Element → Hydrogen at 40 °C, paint, press B, paint again: still Hydrogen 40 °C, not Vacuum.
5. Switch to Power overlay: element cells and element notes fade to 0.3; switch to Gas: gas cells/notes are full alpha, liquid ones faded. Toggle "Show notes" off and on.


Sinetheta:
@coderabbitai review

Sinetheta:
@coderabbitai review

Sinetheta:
@coderabbitai review

Sinetheta:
@coderabbitai full review

PR #236, opened 2026-09-12
feat(auth): add AUTH_MODE=local for a zero-setup dev login

Sinetheta:
## Summary

Every local checkout's `.env` carries WorkOS placeholder values, so
`/api/auth/login|register|send-magic|verify-magic|forgot-password|reset-password`
all fail for a fresh checkout — the only way in was `npm run seed:dev-user`
minting a JWT and pasting `localStorage.setItem(...)` into the browser
console. This adds `AUTH_MODE=local` (set on the devcontainer only, never in
the deploy stack) so a fresh checkout has working logins with no WorkOS keys,
no console paste, and no manual seed step.

- **`app/api/auth-mode.ts`** — `authMode()` reads `AUTH_MODE` (`workos`
  default, or `local`), throws on an unknown value or on `AUTH_MODE=local`
  + `ENV_NAME=production`. Called per-request from the controller, and once
  at boot in `server.ts` to fail fast before the server listens.
- **`app/api/dev-users.ts`** — six dev users (`dev_you` keeps the old
  `seed:dev-user` script's fixed `_id` so old minted tokens/ownership still
  work) sharing one known password (`dev_password`), upserted idempotently
  at boot (`db.ts`, local mode only, fire-and-forget).
- **`auth-controller.ts`** — in local mode, `login`/`register` check the
  stored legacy password directly (existing `setPassword`/`validPassword`
  PBKDF2 machinery, no new hashing scheme) and never call WorkOS;
  `sendMagic`/`verifyMagic`/`forgotPassword`/`resetPassword`/`verifyEmail`
  answer 501 `not_available_in_local_mode`. New `GET /api/auth/mode`
  (public) reports the mode and, in local mode only, the dev-user roster +
  shared password — never a hash/salt.
- **Login/register pages** — fetch `/api/auth/mode` on init; local mode
  shows a "Sign in as…" one-click picker over the dev roster, hides the
  magic-link/forgot-password links, and register logs straight in instead
  of routing to email verification. Small "LOCAL AUTH MODE" banner on both
  pages.
- **`seed-dev-blueprints.ts`** now imports the same dev-user definitions
  from `dev-users.ts` so the social-graph fixture and the boot-time roster
  never drift apart (and picked up fixed `_id`s + the shared password along
  the way); its minted-JWT bypass is kept for an `AUTH_MODE=workos`
  checkout with no real WorkOS keys.
- `.devcontainer/docker-compose.yml` sets `AUTH_MODE: local` on the shared
  node-env block; `.env.sample` documents the var as a comment only (no
  value); docs updated in `CLAUDE.md`, `README.md`, `agent/WORKOS_PLAN.md`.

WorkOS stays the default and the only path in production; nothing about the
WorkOS flow changed.

## Design notes / deviations from the plan

- `ensureDevUsers()` does **not** skip re-hashing on a repeat boot. I tried
  that (skip when `validPassword` already matches), but `validPassword`
  itself pays the same PBKDF2 cost as `setPassword` (~3.4s/user on this
  hardware — crypto-js's real default iteration count, never previously
  exercised by any test, since existing fixtures set `hash`/`salt` directly).
  So "check first" costs the same as "just set it" — the skip bought
  nothing, only complexity, so I reverted to the plan's plain unconditional
  version. Real-world effect: ~20s of hashing once at every container boot,
  fire-and-forget, never blocking the server from serving.
- Added the "LOCAL AUTH MODE" banner to both the login *and* register pages
  (plan said "the auth pages" without enumerating); left the magic-link and
  forgot-password pages themselves unguarded — their links are hidden from
  login, and both already treat every response (including a 501) as "show
  the generic confirmation" for anti-enumeration reasons, so a stray direct
  visit degrades harmlessly rather than crashing.

## Test plan

- [x] `npm run tsc` — clean
- [x] Backend suite — 1127/1127 passing, including new
      `__tests__/api/auth-mode.test.ts` (boot-guard unit tests) and
      `__tests__/api/local-auth.test.ts` (login/register/501s/`/api/auth/mode`/
      `ensureDevUsers` idempotency, plus a workos-mode regression check that
      login never touches `validPassword`)
- [x] `cd frontend && npm test` — 1302 passing / 2 skipped (pre-existing
      skips), including new login/register picker + local-mode specs
- [x] `cd frontend && npm run lint` — clean
- [x] `npm run build` — clean (backend + frontend + lib)
- [ ] **Not yet manually verified** against the running devcontainer stack —
      the manual script from the plan (fresh volume boot, one-click dev-user
      login, form login, register-and-land-logged-in, DB-drop-and-restart,
      `ENV_NAME=production` + `AUTH_MODE=local` refusing to start) still needs
      to be run by hand. PR stays in draft until that's done.

## Update: native PBKDF2 and a two-user boot roster

Seeding at boot was measured at ~20 s of blocked event loop: crypto-js 4.2 defaults PBKDF2 to 250k sha256 iterations in pure JS, ~3.2 s per hash, and the model passed no iteration count. Every local login paid the same.

- `User.setPassword`/`validPassword` now use Node's `crypto.pbkdf2Sync` with the identical parameters. Output is byte-identical to crypto-js (a spec pins a crypto-js-generated fixture hash), so existing legacy hashes keep validating. ~90 ms per hash. `crypto-js` removed from dependencies.
- Boot provisions only `BOOT_DEV_USERS` (`dev_you` admin + `dev_creator_alpha`); the login picker lists just those. The six-account social graph remains a `seed:dev-blueprints` concern. Seed now completes in the same second Mongoose connects.
- The provisioning-window spec gates the seed's first write instead of racing a slow hash; `hooks.ts` clears an inherited `AUTH_MODE` so the suite's WorkOS-mode baseline holds inside the devcontainer.

Verified: `npm test` in the devcontainer, 1131 passing; api restart shows `dev users ensured (2)` within the same second as the Mongoose connection.

## Update: the test suite could wipe the dev database (pre-existing, found here)

Inside the devcontainer the app container carries the dev `DB_URI`, and `hooks.ts` let an inherited `DB_URI` beat `.env.test.local`. So `npm test` in the app container has been dropping the dev database's collections between tests since 2026-09-02; it surfaced when the seeded dev users vanished after a test run.

- `.env.test.local` now overrides the environment (CI has no such file, so its job var still wins over `.env.test`); `scripts/test-db-setup.sh` resolves in the same order.
- `hooks.ts` refuses to start against any database whose name doesn't end in `_test`.

Verified in the container: the suite creates and uses `blueprintnotincluded_test`, the dev users survive a full run, and removing `.env.test.local` makes the guard throw instead of connecting to the dev database.

https://claude.ai/code/session_01VLVT3RXPwDXafb2MLA3VE4


Sinetheta:
@coderabbitai full review

PR #237, opened 2026-09-13
Critter Sensor follow-ups: specific Clear label, catalogue-derived echo keys

dytterud:
Kevin's non-blocking follow-ups from #231, plus the CodeRabbit nitpick that was left open.
No new feature surface — this finishes the picture on the Critter Sensor work.

## What shipped

**1. The Clear button names what it clears** (follow-up 1). `Clear` → `Clear Pressure` /
`Clear Critter count` / `Clear Temperature`, restoring the specificity the old
"Clear threshold" had and matching the `Pressure — (Not set)` row it replaces.
`primarySettingsKey()` already computed the label; the component only exposed it in the
unset direction (`primaryUnsetLabel`), so Clear had no access to it. New `primaryLabel`
getter, one interpolation in the template — the `i18n` attribute is kept, and
`messages.xlf` carries no unit for this component yet, so no translation churn.

**2. The echo key comes from the catalogue, not from `!= THRESHOLD_KEY`** (CodeRabbit).
`clearPrimary()` read *"the primary Key is not `IThresholdSwitch`"* as *"an
`IThresholdSwitch` echo exists"* — true only because the Critter Sensor is today's sole
non-threshold primary. The first prefab with its own primary Key **and** a genuine,
independent `IThresholdSwitch` entry would have silently lost that entry on Clear.

Rather than add a second hand-written branch, `redundantEchoField`'s two `if`s became one
`REDUNDANT_ECHOES` table that both accessors read, and the new
`redundantEchoKeysFor(prefabId, key)` derives the drop list from it. The two can't drift
apart when a second carrier turns up. The panel now loops that answer.

**3. The sandbox confirmation is recorded** (follow-up 2 — **done, no data changes**).
`RangeMax` 64 and all four creatable defaults (`countThreshold` 0; `activateOnGreaterThan`,
`countCritters`, `countEggs` all true) were checked against a sandbox game. The decompile
was right, so nothing moves — but three comments still said "from the decompiled
`[Serialize]` field initializers", and that understated what backs them.

This one matters beyond tidiness: `CREATABLE_SETTINGS`' whole stated policy is *"stays
edit-only-when-present until its real defaults are verified in-game"*, so
`LogicCritterCountSensor` sitting in that list on decompile evidence alone was the single
soft spot in it. It no longer is.

Follow-up 3 (the two unit discrepancies) is now filed as #238.

## Verification

- `npm run build:lib` + `npm run tsc` — clean. The committed `settings-catalog.d.ts` is
  regenerated.
- Backend suite — **1163 passing**, 3 new catalogue tests. 2 failures, both the known
  Windows path-separator assertions in `__tests__/asset-processing/import-scripts.test.ts`,
  which touch nothing in this diff.
- `building-settings.component.spec.ts` — **37 passing**; the two Clear assertions now
  expect the specific label.
- `npm run lint` — clean.
- **In the editor**: a Thermo Sensor's button reads **Clear Temperature**; a Critter
  Sensor's reads **Clear Critter count**, its four rows render the confirmed defaults
  (Threshold 0 / Above / both counts on), and Clear returns it to
  `Critter count — (Not set)`.

### The echo drop, verified against the game

The case that actually motivated the refactor, checked on a round trip through ONI rather
than a fixture. A Critter Sensor configured in-game (threshold 37, "below", count eggs off)
and copied with the BlueprintsV2 mod carries all three keys, the echo mirroring the real
settings:

```json
"buildingData": [
  { "Key": "Switch", "Value": { "switchedOn": true } },
  { "Key": "LogicCritterCountSensor",
    "Value": { "countThreshold": 37, "activateOnGreaterThan": false,
               "countCritters": true, "countEggs": false } },
  { "Key": "IThresholdSwitch",
    "Value": { "Threshold": 37.0, "ActivateAboveThreshold": false } }
]
```

Imported, cleared in the editor, exported, and pasted back into the game:

```json
"buildingData": [
  { "Key": "Switch", "Value": { "switchedOn": true } }
]
```

Both the canonical key and the echo are gone; the stowaway `Switch` is kept. A surviving
echo would have been invisible in the panel — it is suppressed from display *and*
uncounted in "other stored settings" — so the exported file is the only place this could
be checked, and it is also the place where a surviving echo would have done its damage
(the mod's key-apply pass would restore the threshold from it, and Clear would silently
not have cleared).

## Still open from #225

`LogicWattageSensor`, and element sensors.


Sinetheta:
Thanks @dytterud — this closes follow-ups 1 and 2 from #231, and #238 takes care of 3. Merged as-is.

Three things worth naming:

- The self-correction in 67f4153. The first commit moved the over-inference out of the panel into the catalogue, and then you spotted it had landed one layer down — "receives a mirrored field" read as "is a pure copy" — and replaced it with a check derived from `SETTINGS_CATALOG` instead of a comment asserting the invariant. Catching your own fix doing the thing it was fixing is the hard version of this.
- You went to a sandbox for follow-up 2 and then changed nothing but the attribution. `CREATABLE_SETTINGS` is the one table here that synthesizes values into someone's game, and the entry resting on the decompile alone now doesn't. That's the check CI can't do.
- Checking the echo drop through a real round trip, and saying why the export was the only place to check it — a survivor is suppressed from display and uncounted in "other stored settings", so the panel could never have shown you one.

What I checked: all 13 prefabs with a primary settings key, and `redundantEchoKeysFor` agrees with the old `!= IThresholdSwitch` inference on every one — behaviour-identical today, diverging only on the prefab that doesn't exist yet. Transplanting the old inference back in fails your new `LogicSwitch` case immediately. 1166 passing here; the two import-scripts failures you saw are Windows path separators in assertions that predate you, and they're mine to fix.

One non-blocking follow-up: the whole-field guard isn't distinguishable by any test yet — `IThresholdSwitch`'s catalogued fields are exactly the two the table mirrors, so a naive "every Key in the table" passes all three of yours, as your commit message says. A synthetic partial-mirror entry in a test-local table would make it fail now rather than someday, if you want it.

Thanks again.

PR #239, opened 2026-09-13
feat(settings): Wattage and Radbolt sensors are threshold sensors

dytterud:
Kevin's follow-up 2 from #225: *"`LogicWattageSensor` — I believe this is also an
`IThresholdSwitch` carrier in the game code… it should be a one-line addition to the table
with a watts unit."*

He was right. Decompiling to check also turned up a second one, so this adds both.

## The claim that was wrong

`THRESHOLD_SENSORS` excluded these two under "deliberately absent":

> `LogicWattageSensor` and `LogicHEPSensor`. Neither is a confirmed `IThresholdSwitch`
> carrier, and radbolt thresholds demonstrably live on
> `HighEnergyParticleSpawner.particleThreshold` / `HEPBattery.particleThreshold` — different
> keys entirely.

Both halves are wrong. From `ilspycmd` against the installed `Assembly-CSharp.dll`:

```csharp
public class LogicWattageSensor : Switch, ISaveLoadable, IThresholdSwitch, ISim200ms
public class LogicHEPSensor     : Switch, ISaveLoadable, IThresholdSwitch, ISimEveryTick
```

Each backs `Threshold` with a real `[Serialize]` field — `thresholdWattage` /
`thresholdPayload` — rather than aliasing another key the way the Critter Sensor's does, so
these are ordinary threshold sensors, not a repeat of #231's echo problem.

The radbolt sentence conflated the Radbolt **Sensor** with the Radbolt **Generator** and
**Battery**. Those two genuinely do use their own `particleThreshold` key — on different
buildings. They remain unhandled, and the comment now says so without smearing the sensor.

Being absent wasn't cosmetic: with no entry these sensors showed a bare unitless `Threshold`
float, got no `Switch` suppression (so the stowaway sampled output was offered as an
editable "On" row), and couldn't create the key from scratch.

## Values

| | `LogicWattageSensor` | `LogicHEPSensor` |
|---|---|---|
| Stored field | `thresholdWattage` | `thresholdPayload` |
| `RangeMin` / `RangeMax` | 0 / **75000** — `1.5f * Wire.GetMaxWattageAsFloat(Max50000)` | 0 / **500** — flat `maxPayload` |
| Unit | watts 1:1 — `Sim200ms` compares `thresholdWattage` straight against `circuitManager.GetWattsUsedByCircuit()` | radbolt count 1:1 |
| Label / suffix | `Wattage` / `W` | `Radbolt` / `radbolts` |

Neither converts (`displayScale: 1`). Labels follow the table's existing convention of
shortening the game's side-screen value names ("Wattage Reading", "Radbolt Reading"), the
same way "Ambient Pressure" became `Pressure`.

Two table entries; everything else falls out of them — `Switch` suppression,
`IThresholdSwitch` in `CREATABLE_SETTINGS` via the existing loop, and `primarySettingsKey`
for the unset row.

## Verification

- `npm run build:lib` + `npm run tsc` — clean. No `.d.ts` churn (the `Record<string, …>`
  signature is unchanged).
- Backend suite — **1164 passing**, 4 new tests. The 2 failures are the known Windows
  path-separator assertions in `import-scripts.test.ts`, untouched by this diff.
- `building-settings.component.spec.ts` — 37 passing (unchanged; the panel is
  prefab-agnostic).
- `npm run lint` — clean.

The new tests include the "confirmed carriers and nothing unverified" list — that assertion
is the record of what has been checked, so it is updated deliberately — plus coverage that
neither entry introduces a conversion, and that `Switch` is suppressed on both.

Live in the editor, both sensors:

| | Wattage Sensor | Radbolt Sensor |
|---|---|---|
| Unset row | `Wattage — (Not set)` + Set | `Radbolt — (Not set)` + Set |
| After Set | `1000`, suffix `W`, min 0 / max 75000 / step 1 | `10`, suffix `radbolts`, min 0 / max 500 |
| Stray "On" checkbox | none | none |
| "other stored settings" count | none | none |

### Confirmed against the game, both directions

The units are not taken on the decompile's word. A blueprint was exported through
`Blueprint.toBniBlueprint` — the same path a download from the site uses — pasted into the
game, read on each side screen, then **copied back out with the mod**:

| Building | Written | Side screen | Copied back |
|---|---|---|---|
| Wattage Sensor | `1234` / Below | 1234 W — **watts 1:1 confirmed** | `1234.0` / `false` |
| Radbolt Sensor | `137` / Below | 137 radbolts | `137.0` / `false` |
| Thermo Sensor (control) | `293.15` / Above | 20 °C | `293.15` / `true` |

Every threshold and direction survives the full loop unchanged, as do the positions and the
`selected_elements` material hash. Values were deliberately non-round so none could be
mistaken for a game default, and two were set to Below so the direction field was exercised
too. The Thermo Sensor is an already-confirmed carrier included as a control: had it failed
as well, the fault would have been in the export rather than in the new entries.

The copy-back also shows the game attaching **`Switch: {switchedOn: true}` to all three**
sensors, which it did not have on the way in. That is the stowaway sampled output — these
prefabs extend `Switch`, so the mod's `Switch` handler matches them — and it is the concrete
reason to put these two *in* the table rather than leave them out: the table entry is what
makes `resolveSettingDescriptors` suppress that key instead of rendering it as an editable
"On" checkbox the user would reasonably think meant something.

## Notes

- Independent of #237 — that branch touches `settings-catalog.ts`; this one touches
  `threshold-sensors.ts`. Once #237 lands, these sensors' buttons read **Clear Wattage** /
  **Clear Radbolt** for free.
- Still open from #225: element sensors.


Sinetheta:
Thanks @dytterud — this is follow-up 2 from #225 done properly, and it turned up the second sensor I'd missed as well.

The part I want to name specifically: the claim you demolished here was your own, from #225, and you went and tested it rather than treating a settled comment as settled. Both halves were wrong for different reasons and you say which is which — the carrier claim, and the radbolt sentence that had conflated the Radbolt *Sensor* with the Generator and the Battery. The corrected comment stays in the file naming what's still unhandled, so the next reader inherits the evidence rather than the conclusion.

The in-game loop is the other one. Non-round values, two set to Below so the direction field was exercised, and a Thermo Sensor carried through as a control so a failure in the export couldn't be mistaken for a failure in the new entries. That's the check that matters most here and it's not one CI can do.

I checked the decompile independently against a public mirror: both class declarations, `maxWattage = 1.5f * Wire.GetMaxWattageAsFloat(Max50000)` with `Max50000` returning `50000f`, `maxPayload = 500f`, and the `Sim200ms` comparison of `thresholdWattage` straight against `GetWattsUsedByCircuit()` that makes watts 1:1. Your four specs are load-bearing — pull the two table entries and exactly those four fail. Merged as-is.

Non-blocking follow-ups, if you want them:

1. `HighEnergyParticleSpawner` and `HEPBattery` — now precisely named as unhandled. They're `ISingleSliderControl` rather than `IThresholdSwitch`, so it's a new key rather than a table row. An issue would do.
2. The three pulled switches (`PressureSwitchGas`/`PressureSwitchLiquid`/`TemperatureControlledSwitch`) were waiting on a debug build-menu check. You now have both a decompile and a working in-game round-trip, which is more than that note was asking for.
3. Element sensors, still open from #225.

Thanks again.

PR #240, opened 2026-09-15
fix(editor): keep the current overlay when the picked item declares none

dytterud:
> [!NOTE]
> **Ready for review.** Every manual check in the editor passes, and @Sinetheta has settled the one open product question: the element brush stays as it is.
>
> **Merge #241 first, or together with this.** Not a code dependency -- different files, both merge clean -- but the B-key and element-brush paths here run into a crash that `aa0bb8d1` put on master, and #241 is what fixes it. Landing this one alone makes that crash look like this PR's doing.
>
> Changes what the *view* does when you pick a building, so the editor pass below is the real verification; the specs only pin the rule.

## Manual testing so far

Tested in the dev editor by @dytterud, against a local branch merging this PR with #241 (needed because a tool switch crashes without it):

- ✅ Automation overlay → pick a **Lamp** / **Apothecary** → stays in Automation. *The behaviour this PR exists for.*
- ✅ → pick a **Gas Pump** → Gas; **Battery** → Power. The follow still works.
- ✅ **B** over a placed Lamp (stays) and a Gas Pump (switches); same via the select tool, so all three `changeItem` / `same-item-collection` entry paths agree.
- ✅ **Blank Canvas** (Decor), **Ceiling Light** (Light), **Critter Drop-Off** (Rooms) in Automation → stay put. These are the 64 buildings whose overlay this site does not render; the divergence from ONI is not jarring in practice.

Replaces #232, in the opposite direction. @Sinetheta answered the product question that PR asked, and the answer closed it: ONI's overlay-follow is unconditional and deliberate, so gating it on "the item is already visible" moves us away from the game. #232 is closed. This is the gate the same decompile turned up instead.

## The bug

Picking a Bed or a Ladder from the build menu while you are in the Automation overlay drops the view to Base. That is the original report's actual complaint, and neither master nor #232 fixed it.

`InterfaceTool.OnActivateTool`, from [Kupie/ONI_Decomp](https://github.com/Kupie/ONI_Decomp) at `ChangeList = 744825` — the same build as our database baseline, `U59-744825-SCRPAN`:

```cs
if (OverlayScreen.Instance != null && this.viewMode != OverlayModes.None.ID
    && OverlayScreen.Instance.mode != this.viewMode)
{
    OverlayScreen.Instance.ToggleOverlay(this.viewMode, true);
}
```

`BuildingDef.ViewMode` defaults to `OverlayModes.None.ID` — Bed and Ladder set nothing, Gas Pump sets `GasConduits`. `BuildTool.Activate` sets `this.viewMode = def.ViewMode`, and Copy Building routes through `PlanScreen.CopyBuildingOrder` → `OnSelectBuilding` → the same call, so the game shares one path there exactly as `changeItem` does here. When the def declares no viewMode, the game leaves your overlay alone.

## The change

`convert-export-2024.ts`'s `overlayFromViewMode` maps a missing viewMode to `Overlay.Base`, so `item.overlay === Overlay.Base` **is** ONI's `ViewMode == None` test:

```ts
if (this.overlay_ !== Overlay.None && item.overlay === Overlay.Base) return;
this.overlay = item.overlay;
```

The follow stays unconditional for the 243 buildings that declare an overlay we render. Nothing about #232's visibility test survives.

### The `Overlay.None` guard — and a correction to what we both assumed

The guard is in, but **the reason we both gave for it does not hold**, so the comment and the spec now say something narrower (`ea6556ee`).

We assumed the editor boots at `Overlay.None`, that `oniItemsLoaded()` seeding a Tile is what leaves it, and therefore that a bare early return would boot the editor with no overlay at all. Instrumenting `setOverlayForItem` in the dev editor shows the very first call is already gated:

```
[DBG setOverlay] Tile itemOverlay= 0 camera= 0 gated= true
```

`component-canvas.component.ts` sets `cameraService.overlay = Overlay.Base` on init, before `fetchDatabase()` resolves and `oniItemsLoaded()` runs. The camera has left `None` well before the boot Tile reaches this method, and the Tile's own call then changes nothing either way.

The guard and its spec stay, on the ground that actually holds: `CameraService`'s constructor still defaults to `None`, nothing enforces that initialisation order, and `Overlay.None` is genuinely not a neutral state — nothing is primary or secondary there, so `BlueprintItem.cameraChanged` renders everything at 0.3 alpha on bare `zIndex`, which also changes select-tool frontmost resolution since it picks max `depth`. Handling that beats assuming it away. But it is not load-bearing for the boot sequence, and this PR should not have said it was.

## Two deliberate divergences from the decompile

Counted from `assets/database/database-2024.json`, where `viewMode` is already stored as an `Overlay` value:

| `item.overlay` after load | buildings | ONI |
|---|---|---|
| Base, `viewMode` absent | 180 | does not switch — we now match |
| Base, via `getRealOverlay` collapsing Oxygen/Decor/Light/Temperature/Rooms/Radiation | 64 | switches, to an overlay this site does not render |
| Power / Liquid / Gas / Automation / Conveyor | 243 | switches — unchanged here |

**1. Those 64.** We cannot follow the game for them; there is no such overlay here. Today they yank you to Base, after this they leave you where you are. Better, but not parity, and worth naming rather than glossing. `CreatureDeliveryPoint` (`viewMode` 10, Rooms) is pinned in the contract spec as the witness.

**2. The element brush — an open question for you.** `paintElement()` routes through `BuildTool.changeItem` like any building, and the synthesized `Element` OniItem (`OniItem.load`) has `objectLayer = 0`, `zIndex = ZIndex.GasFront`, `overlay = Overlay.Base`. So it currently pulls you to Base, and after this it will not: painting an element while in Automation paints into a view where elements sit at 0.3 alpha.

I did *not* add an `isElement` carve-out. It is the same behaviour class as placing a Bed in Automation, which ONI accepts, and inventing an exemption you did not ask for seemed worse than putting it in front of you. It is pinned by a spec so it is visible rather than discovered. **If you want the element brush exempted, say so and it is one line.**

## Verification

- `npm run build:lib && npm run tsc` clean.
- Backend suite: **1177 passing, 2 failing** — the two `import-scripts` Windows path-separator assertions you said are yours; they predate this branch and fail on master here too.
- The 10 new specs bite, checked by reverting the source:
  - against master's unconditional assignment → the 4 "stays put" specs fail (the three overlays plus the element case);
  - against the *bare* `if (item.overlay === Overlay.Base) return;` → the `Overlay.None` spec fails (a unit-level property of the method, not the editor bootstrap — see the correction above).
- The contract spec pins both fake shapes to real database buildings (`Tile`/`Bed`/`Ladder` → Base; `GasPump`/`LogicSwitch`/`Battery`/`LogicWire` → their own), asserts the `getRealOverlay` collapse, and asserts no `OniItem` can ever carry `Overlay.None` — so the None branch only ever fires for the camera's own initial state. Kept from your `34496853`; the assertions had to invert but the idea is why this block tests the code rather than a model of it, so you are on the commit.

`build-tool.spec.ts:166` and `same-item-collection.spec.ts:151` mock `setOverlayForItem`, so the frontend suite is untouched.

- ✅ **Boot.** After a reload the Overlay menu reports **Buildings** (`Overlay.Base`) checked, and placed tiles render fully opaque — the camera is not stranded at `Overlay.None`.
- ✅ **Element brush in Automation.** Painting works and the view stays in Automation. Side by side: in Automation the painted water is pale and washed out and the tile outlines are grey; switching to Buildings makes the same cells saturated blue with solid outlines. The difference is clearly visible, but the painted water stays **legible** — you can see what you painted and where. The brush preview under the cursor is likewise faint but visible.

## The one open decision

Whether that dimmed element brush is acceptable, or whether the element brush alone should keep its old pull-to-Base. It is one line either way.

Argument for leaving it as-is (my inclination):

- It is exactly how every other building behaves in a non-matching overlay, so the element brush is not made a special case.
- It matches the game, which does not switch for `ViewMode == None`.
- An `isElement` carve-out would make the element brush the single exception to the rule this PR establishes.

Argument against:

- The element brush is a *painting* tool, where seeing your work matters more than when placing one building, and it is the only tool where you might lay down many cells in a row.

@Sinetheta — your call.

#241 should land first, or at least alongside — a tool switch on this branch alone hits that crash.

## Still open from the same report

Untouched here: automation buildings rendering behind automation wire (a depth-band issue in `BlueprintItem.cameraChanged` — Building-layer devices land at `zIndex + 50` = 69 while the wire sits at `zIndex + 100` = 113), and missing port markers on logic gates.


Sinetheta:
Thanks @dytterud, this is looking good. I read it through and re-ran your load-bearing check from my side (four specs fail against master, the None bootstrap spec fails against the bare gate, full suite green at 1179), and your 180 / 64 / 243 split matches my own count of the shipped database. I am leaving it in draft this time; it is yours to flip when the editor pass is done, and I'll take it from there.

Two things while you are in there. On the element brush, leave it as you have it: it is the same class as placing a Bed in Automation, the game accepts that, and the spec you pinned means any later change is on purpose. Do note what it actually looks like when you paint in a non-Base overlay, since neither of us has seen it. And the `Overlay.None` comment is the right kind, explaining it as a rendering state rather than a boot quirk; that is what stops the next person from simplifying the line.

No rush.

dytterud:
Thanks — taking the element brush as settled, leaving it as it is, and the spec stays so a later change has to be deliberate.

On what it actually looks like, since you asked and neither of us had seen it. Painting water in the Automation overlay: the placed cells come out pale and washed-out, the tile outlines beside them grey, and the brush preview under the cursor is faint but visible. Switching to Buildings makes the same cells saturated blue with solid outlines. The difference is obvious side by side — but the dimmed version stays *legible*: you can see what you painted and where. It reads like every other building in a non-matching overlay, which is the argument for leaving it. I'd expected worse when I flagged it.

One correction, on the part you singled out. You read the `Overlay.None` comment as it stood at 04:48; `ea6556ee` rewrote it 14 minutes later, because the premise was wrong. The comment claimed the editor boots at `Overlay.None` and that `oniItemsLoaded()` seeding a Tile is what leaves it. It doesn't. Instrumenting the call in the dev editor shows the very first one is already gated:

```
[DBG setOverlay] Tile itemOverlay= 0 camera= 0 gated= true
```

`component-canvas.component.ts` sets `cameraService.overlay = Overlay.Base` on init, before `fetchDatabase()` resolves, so the camera has left `None` long before the boot Tile reaches the method. The guard stays — the constructor still defaults to `None`, nothing enforces that ordering, and `None` really is the 0.3-alpha-on-bare-`zIndex` state — but it isn't load-bearing for the boot sequence and the PR shouldn't have said it was. The rendering-state half you liked is now the whole justification rather than the second sentence.

Editor pass is done: Lamp/Apothecary stay in Automation; Gas Pump → Gas and Battery → Power still follow; B-key and select tool agree with the build menu on both; Blank Canvas / Ceiling Light / Critter Drop-Off stay put and don't feel jarring; reload boots into Base at full opacity.

One dependency worth knowing about: #241. Testing the B-key and element-brush paths here means switching tools, and on master that crashes — `switchFrom()` destroyed the build brush and left it in the field, which `aa0bb8d1`'s new id check then kept and drew. Not caused by this PR; it reproduces with this gate reverted.

Sinetheta:
Thanks @dytterud. #241 is in, so this one goes next.

The editor pass covers everything I wanted to see, and the element-brush observation is the answer to the question I asked: dimmed but legible is the same thing a Bed gets in Automation, so the pinned spec stands and any later exemption is a deliberate change. The corrected `Overlay.None` comment is better than the one I praised at 04:48, and the correction is the right shape: you instrumented the boot sequence rather than reasoning about it, found the premise wrong, and kept the guard on the ground that actually holds, that `None` is a rendering state the method must not strand a camera in. That is the version that stops the next person simplifying the line.

The 180 / 64 / 243 split still matches my count, and the contract spec against the shipped database is what makes the fakes trustworthy. Merged as-is.

One non-blocking follow-up if you want it: the 64 buildings whose overlay this site does not render (`CreatureDeliveryPoint` is your witness) now stay put where the game would switch. If any of those overlays ever gets rendered here, `getRealOverlay` is the one place to widen, and the contract spec will tell you the moment it happens. No action now; just noting where the seam is.

Thanks again.

PR #241, opened 2026-09-15
fix(editor): stop destroying the build brush when switching tools

dytterud:
> [!NOTE]
> **Ready for review.** Fixes a regression `aa0bb8d1` introduced three days ago, live on master now: switching away from the build tool and back leaves a destroyed brush in the draw loop, which throws inside the PIXI ticker and stops the editor repainting entirely.
>
> Verified: backend 1173 / frontend 1330, CI green on all six jobs; ten new specs, nine failing against master; a six-case tool-switch pass in the dev editor; and an end-to-end save -> render -> Discover thumbnail round trip, since `destroy()` is shared with the backend preview renderer.
>
> **Merge this before or together with #240.** Not a code dependency -- they touch different files and both merge clean -- but #240's B-key and element-brush paths run straight into this crash, so landing that one first makes it look like #240's fault.

Found while manually testing #240; it is not caused by that PR — it reproduces with #240's change reverted.

## Hit incidentally, not just by the deliberate repro

Worth adding, because it changes how likely this is to reach users. While building an unrelated test blueprint on another branch (#244, which is based on master and so does not have this fix), the editor died again — same signature, no attempt to trigger it:

```
TypeError: Cannot read properties of null (reading 'position')
    at r.set ... at t.drawPixi ... at t.draw ... at t.drawAll ... at _tick
```

That was ordinary editing: placing buildings, switching between the build and select tools, and **Blueprint → New**. Nothing exotic. The render loop stopped, later placements silently went nowhere, and the canvas simply stopped responding — no error surfaced in the UI. Merging this fix into that branch locally made it go away, which is what made #244's own testing possible at all.

So this is not an edge case that needs a specific sequence to reach. It takes a few minutes of normal use.

## Repro

Editor → **Automation overlay → element brush → Select tool → press B.** The editor stops repainting entirely: no build preview, and the overlay appears not to change either. Console:

```
Uncaught TypeError: Cannot read properties of null (reading 'position')
    at t.drawPixi ... at t.draw ... at t.draw ... at t.drawAll ... at i.update
```

Not reproducible on production, because production predates `aa0bb8d1` (see below).

## Cause

`BuildTool.switchFrom()` destroys `templateItemToBuild` but leaves it in the field — unlike `destroy()` a few lines above, which nulls it. A destroyed `BlueprintItem` keeps `containerCreated = true` and a destroyed PIXI container, so `drawPixi`'s `this.container.x = …` hits a null `transform`. That throw happens inside the PIXI ticker, which is why one bad brush stops *every* subsequent frame — hence "no preview and the overlay never changed", when in fact nothing at all is repainting.

The stale field dates to the initial commit, but nothing could reach the corpse until `aa0bb8d1` ("editable element brush, B-key sampling, overlay-aware fading", 3 days ago). Before it, re-entering the build tool always rebuilt the brush:

```ts
if (toolType == ToolType.build) this.uiItemChanged();
```

`aa0bb8d1` made that conditional (`ensureBuildItem`), for a deliberate and good reason — from its own commit message:

> Re-entering the build tool (toolChanged) only rebuilds the brush when the incoming item actually differs, fixing "B" silently resetting an already-configured element brush back to Vacuum.

The guard only compares `oniItem.id`, though, so a *destroyed* brush whose id still matched was kept — and then drawn. Two paths reach `toolChanged(build)` and they behave very differently:

| path | `switchFrom` called? | result on master |
|---|---|---|
| **B** while already in the build tool | no — `changeTool` skips the tool you are on | brush preserved, `aa0bb8d1` working as intended |
| switch to another tool, then back | yes — brush destroyed, field left populated | destroyed brush kept by the id check, then drawn → crash |

So `aa0bb8d1` delivered its feature on the first path and crashed on the second. This PR makes the second path work the way the first already does, rather than removing the guard.

Instrumented trace of the repro:

```
[DBG changeItem] Element old= Tile oldDestroyed= false
[DBG switchFrom] destroying Element
[DBG BuildTool.draw on destroyed template] Element
[DBG DESTROYED draw] Element buildCandidate= true containerCreated= true containerNull= false
Uncaught TypeError: Cannot read properties of null (reading 'position')
```

## The fix

**`switchFrom()` hides the brush instead of destroying it** (`setInvisible()`, which is what `mouseOut()` already uses to park it off-screen). `draw()` runs only while the build tool is the current tool, so an inactive brush costs a hidden container and nothing else. `changeItem()` still destroys the outgoing brush on a real item change — the one place a brush genuinely becomes garbage.

With the brush alive, `ensureBuildItem`'s id check works as `aa0bb8d1` designed it and hands the same configured item back. So this doesn't just stop the crash, it delivers the behaviour `aa0bb8d1` was after and missed on this path:

| | before B | after B |
|---|---|---|
| prod (pre-`aa0bb8d1`) | Water brush | **Vacuum** brush |
| master (with `aa0bb8d1`) | Water brush | **TypeError, render loop dead** |
| this PR | Water brush | **Water** brush, paints |

Two guards stay as backstops. Neither should fire now; both are cheap, and the failure they prevent takes down the whole editor:

- `BuildTool.draw()` skips a null or destroyed brush rather than throwing inside the PIXI ticker.
- `ensureBuildItem()` rebuilds rather than keeping a destroyed brush.

### Two approaches tried and discarded

1. **Null the field in `switchFrom()`** to match `destroy()`. Looks obvious, is wrong: `hover`/`updateBuildCandidateResult`, `leftClick` and `mouseDown` all read `templateItemToBuild` unguarded, and the live repro then produced `Cannot read properties of null (reading 'buildCandidateResult')` — one crash traded for another.
2. **Keep destroying, and rebuild the destroyed brush in `ensureBuildItem`.** Stops the throw, but lands back on prod's behaviour: every tool switch resets a configured element brush to Vacuum, which is the exact thing `aa0bb8d1` set out to fix. Kevin's commit message names it directly. Shipping that would have been a crash fix that quietly undid a feature.

## Verification

- Frontend suite **1327 passing**, 2 skipped (was 1323; +4 new). Backend suite **1171 passing** (+4 new lifecycle specs). CI green on all six jobs.
- Ten new specs in total; nine fail against master.
- New specs, all failing against master:
  - `switchFrom` hides the brush / does not destroy it / does not throw with no brush
  - `draw` does not draw a destroyed brush, and does not draw when there is none
  - `toolChanged` rebuilds when the matching template item has been destroyed
- **Driven through the dev editor by me** (scripted browser, not a person): the repro runs clean, no console errors, the panel still reads `Water` afterwards and the canvas shows a water preview rather than the 🚫 no-build marker. Checked with #240's gate both applied and reverted — same crash before, same fix after.

## A regression this branch caused, found in manual testing

Worth recording, because no spec caught it and no spec could have. After the `setInvisible` change, picking something to build and then **right-clicking** into the select tool left the ghost brush parked on the canvas at the last cursor position, until a new pick destroyed it.

`setInvisible()` only sets `position = (-99999, -99999)`, and that position reaches the PIXI container in `drawPixi` — which stops being called the instant `ToolService.currentTool` changes. So the last drawn frame stays on screen. Destroying used to handle it as a side effect: a destroyed container leaves the stage.

Fixed in `41654dfc` with `BlueprintItem.setDrawnVisible(visible)`, which toggles the container **and** the utility sprites — those live on the camera's container rather than the item's, so a Gas Pump brush would otherwise have left its port icons behind too. `switchFrom` hides, `switchTo` shows.

The specs asserted `setInvisible` was called. That was true, and still wrong: the bug lived in what the renderer did afterwards.

## Manual testing so far

Tested in the dev editor by @dytterud, on a local branch merging this PR with #240:

- ✅ Right-click out of the build tool leaves no ghost brush.

Tool-switch shakeout, driven through a scripted browser against the same merged branch (so: real editor, real renderer, but not a person's hands):

| case | result |
|---|---|
| **Gas Pump** brush → right-click | ghost and all three utility-port sprites gone. Chosen deliberately: Gas Pump has Power, Gas *and* Automation ports, and those sprites live on the camera's container rather than the item's, so they are the ones most likely to be orphaned |
| building → **Escape** | clean |
| building → **toolbar** select | clean |
| element brush set to **Water** → `V` → `B` | comes back as Water, not Vacuum — `aa0bb8d1`'s behaviour intact |
| **wire** run placed → Escape | placed wires remain (blueprint content), brush ghost goes. Wires take `BlueprintItemWire.drawPixi`, patched separately |
| **15 rapid mixed switches**, including the unbound toolbar tools (planning, scissors, notes, terrain) | no errors, canvas still repainting, brush preserved throughout |

Zero console errors across all of it.

## Still needs human testing

Everything above that is marked scripted was driven by tool calls, not by a person using the editor. The failure modes on this branch — a ghost sprite, a stalled canvas — are ones a person notices instantly and an assertion does not, so a few minutes of ordinary editing is still worth more than the table above.

**The backend preview round trip is now covered.** `destroy()` is shared with the preview renderer, which no test suite exercises, so it was checked end to end in the dev container:

- Built a blueprint in the editor, saved and published it as a normal user.
- The API rendered all three variants at save time — `card.webp` 11.5 KB, `hero.webp` 27 KB, `og.png` 171 KB, `renderVersion: 2`. Real content, not blank.
- `backfill-preview-images` then reported it `fresh` (correctly skipping an up-to-date render).
- The thumbnail displays correctly on Discover, showing the placed tiles.
- `preview-render-worker --smoke` also passes independently: 19% pixel coverage against its 5% floor.

Nothing in the renderer is affected by the change, and the reason is structural rather than lucky: both `pixi-node-util` and `preview-render-worker` tear down the *camera's* container with `{children: true}` and never call `BlueprintItem.destroy()`, so `destroyed` is never set and neither the early return nor the new `destroy()` body is reached.
- **Blueprint preview images still render correctly**, since the `destroy()` change is in `lib/` and shared with the backend renderer. Open a few saved blueprints and check their thumbnails, and ideally run `npm run backfill-previews -- --dry-run` against something real.
- **Any** tool switch away from build and back, with a regular building as well as the element brush — including via right-click and Escape, which route to `changeTool(select)` and so also hit `switchFrom`.
- That a configured brush (element, temperature, material, orientation) survives a round trip, and that a genuine item change still swaps it.
- That nothing leaks or double-draws now that the brush is retained rather than destroyed — a long editing session with many tool switches is the honest test.
- That pressing **B** with something selected, and with something under the cursor, still samples correctly (`changeToBuildToolFromSelection`).

## Sweep of the other tools

`BuildTool` was the only offender. No other tool destroys anything in `switchFrom` — they all hide, which is what this PR now makes `BuildTool` do:

| tool | `switchFrom` does |
|---|---|
| notes, planning, terrain | `hoverTile = null`, `preview.visible = false` |
| scissors | clears `startTile` / `startFloat` / `direction`, `readyIcon.visible = false` |
| select | `deselectAll()`, `hideDimensionsText()` |
| element-report, same-item-collection | no `switchFrom` |

Widening to the underlying pattern — destroy something, keep the reference, draw it later — there are four `destroy()` call sites across the tools, drawing and blueprint layers, and the other three are correct:

- `build-tool.ts` `destroy()` — nulls the field afterwards.
- `build-tool.ts` `changeItem()` — destroys the outgoing brush and replaces it on the next line; the `draw()` backstop covers the gap.
- `blueprint.ts` `destroyBlueprintItem()` — splices the item out of `blueprintItems` **before** destroying it, so `drawAll` cannot reach it.

No other draw path can reach a destroyed item: `SelectTool.draw` only draws the selection box, and `same-item-collection` has no `draw`.

## Hardening `BlueprintItem.destroy()`

The reason this bug was a dead editor rather than a missing sprite: `destroy()` destroyed the PIXI container and the utility sprites but left `containerCreated = true`, the container reference, and the sprite array in place. A destroyed item was indistinguishable from a live one to every drawing path, so any stale reference produced a throw rather than a no-op — inside the ticker, taking every later frame with it.

`destroy()` now drops the container, clears `containerCreated`, and empties `utilitySprites`; `drawPixi` returns early when `destroyed`. **Both are needed together**: clearing the container alone would make the next draw build a fresh one and silently *resurrect* an item that is meant to be gone, and the early return alone would leave the PIXI references held. `BlueprintItemWire.drawPixi` bails too — its pipe circles draw through `pixiUtil` rather than `this.container`, so the base-class return doesn't cover them.

Callers are still expected to drop their references. This is the safety net for the ones that don't, and it turns a class of whole-editor failures into a skipped draw.

This part touches `lib/`, so it affects the backend preview renderer as well. Both `pixi-node-util` and `preview-render-worker` draw live items and never destroy them, so the early return is inert there — but that is reasoning, not a measurement, and the rendered-output check is on the list below.


Sinetheta:
Thanks @dytterud, and sorry for the silence; I was travelling and you had this ready four days ago. Taking this one first, as you asked.

The regression is mine: `aa0bb8d1` narrowed the rebuild to "only when the item differs" and never asked whether the item it was keeping was still alive. Your table of the two paths into `toolChanged(build)` is the clearest statement of what I broke, and it is also why the fix is right. Hiding in `switchFrom` instead of destroying is what makes the id check in `ensureBuildItem` mean what I intended, so this lands the configured-brush-survives-a-tool-switch behaviour rather than just stopping the throw. The two approaches you tried and discarded, and why each failed, is exactly the reasoning I want in a PR description; nulling the field looked obvious and you showed it trades one crash for another.

Two other things worth naming. The ghost brush you found in manual testing, and the `setDrawnVisible` fix that hides the utility sprites too because they live on the camera's container, is a bug no spec could have caught and the Gas Pump case is the right one to check. And the preview-renderer round trip through save, render and Discover, since `destroy()` is shared with the backend: that is the check that matters most for a `lib/` change and it is not one CI can do.

From my side: both suites green here (backend 1175, frontend 1330), lint clean, and putting master's `switchFrom` and `ensureBuildItem` back fails five of the new frontend specs, while master's `destroy()` fails three of the lifecycle specs. Merged as-is; any cleanup I care about I'll do myself in follow-ups, that's the house policy here.

Thanks again.

Issue #242, opened 2026-09-15
Build menu offers 12 deprecated buildings that cannot be built in game


dytterud:
The build menu offers 12 buildings that cannot be built in the game, because the 2024 export emits no `Deprecated` flag for the converter to filter on.

## What the game does

`BuildingDef` gates menu visibility on the flag:

```cs
return !this.Deprecated && (!this.DebugOnly || Game.Instance.DebugOnlyBuildingsAllowed);
```

Note the asymmetry — `DebugOnly` buildings are revealed by debug mode, `Deprecated` ones are excluded in *every* mode. A PLANORDER entry in `TUNING/BUILDINGS.cs` is not menu reachability; several of these still have one.

24 Configs set `buildingDef.Deprecated = true` (decompile: `Kupie/ONI_Decomp`, ChangeList 744825 — the same build as our `U59-744825-SCRPAN` baseline). 23 of them are in `database-2024.json`, and 12 are reachable in our build menu:

| prefab | name |
|---|---|
| `AstronautTrainingCenter` | Space Cadet Centrifuge |
| `CreatureDeliveryPoint` | Critter Drop-Off (superseded by `CritterDropOff`) |
| `GasConduitOverflow` | Gas Overflow Valve |
| `GasConduitPreferentialFlow` | Priority Gas Flow |
| `LiquidConduitOverflow` | Liquid Overflow Valve |
| `LiquidConduitPreferentialFlow` | Priority Liquid Flow |
| `LiquidCooledFan` | Hydrofan |
| `PressureSwitchGas` | Atmo Switch |
| `PressureSwitchLiquid` | Hydro Switch |
| `RoleStation` | Skills Board |
| `SteamTurbine` | **[DEPRECATED] Steam Turbine** (superseded by `SteamTurbine2`) |
| `TemperatureControlledSwitch` | Thermo Switch |

`SteamTurbine` carries the marker in its own display name, which is the clearest corroboration that this class of building is not meant to be offered.

Spot-checked `AstronautTrainingCenterConfig`, `CreatureDeliveryPointConfig`, `GasConduitOverflowConfig` and `LiquidCooledFanConfig`: the assignment is unconditional in each, not guarded by DLC or game version.

## Why it matters

A blueprint containing any of these is unbuildable — the mod's apply pass has no prefab to place. It is also a quiet correctness problem for anything derived from building ids: `requiredDlcs`, `modded` detection and category scoring all treat these as ordinary base-game buildings.

[...]

PR #243, opened 2026-09-15
docs(settings): Atmo/Hydro/Thermo Switch are deprecated, not unverified

dytterud:
Closes follow-up 2 from #239 — the one asking whether Atmo/Hydro/Thermo Switch are placeable at all. They are not. Comment-only change; the table is untouched.

## The answer

All three Configs set the flag outright:

```cs
buildingDef.Deprecated = true;
```

and `BuildingDef` gates menu visibility on exactly that:

```cs
return !this.Deprecated && (!this.DebugOnly || Game.Instance.DebugOnlyBuildingsAllowed);
```

(`Kupie/ONI_Decomp` at ChangeList 744825 — the same build as our `U59-744825-SCRPAN` database baseline.)

So @Sinetheta's original reasoning was right: pre-Automation-Update simple threshold switches, superseded by Sensor + Logic Gate. The wiki gap and the missing Electrical menu entry were true negatives.

## Two things worth recording, which is why this is a comment and not just a closed question

**The test the note asked for could never have worked.** It said to confirm "in a debug/sandbox build menu (which shows disabled content)". Look at the asymmetry in that expression: `DebugOnly` buildings are revealed by debug mode, `Deprecated` ones are excluded in *every* mode. Anyone following that instruction would have found nothing and learned nothing from it.

**A PLANORDER entry is not menu reachability.** My first pass at this went the other way and restored all three to the table with units, ranges and passing specs, because `TUNING/BUILDINGS.cs` lists them in the Power category's `PlanScreen.PlanInfo` under subcategory `switches`, right after `SwitchConfig.ID`:

```cs
SwitchConfig.ID,
LogicPowerRelayConfig.ID,
TemperatureControlledSwitchConfig.ID,
PressureSwitchLiquidConfig.ID,
PressureSwitchGasConfig.ID
```

That entry is real but inert, and `Deprecated` is checked separately. @dytterud caught it with "it is not under power!", which sent me to the Configs. The note now records the trap so the next reader doesn't repeat the inference.

The units are recorded too, in case the buildings ever come back: kg stored / g shown for gas (rangeMax 2), kg for liquid (rangeMax 2000), Kelvin stored / Celsius shown (maxTemp 573.15). All three do carry a genuine `IThresholdSwitch` with its own `[Serialize]` backing field — being deprecated is the only reason to leave them out.

## Found on the way

The site's build menu offers all three, because `buildMenuItems` carries an Electrical entry for each and the export emits no deprecated flag to filter on. That turned out to be 12 buildings, not 3 — filed as #242, blocked upstream on the export side.

## Verification

Comment-only. `npm run build:lib` clean, the 15 threshold specs pass unchanged, backend suite green apart from the two pre-existing Windows path-separator failures in `import-scripts` that pass on Linux CI.


dytterud:
Closing — the evidence is all recorded in #242 (the `BuildingDef` gate, the `DebugOnly` vs `Deprecated` asymmetry, and the 12 affected buildings), so whoever picks that up can update this comment as part of the fix rather than having a doc change land ahead of it.

PR #244, opened 2026-09-15
feat(editor): edit element-sensor / filter element (buildingData)

dytterud:
> [!NOTE]
> **Ready for review.** Closes follow-up 3 from #239 (open since #225). Verified in the editor end to end *and* in the game — see below.
>
> **Merge #241 first.** Not a code dependency, but #244 is based on master and so carries the build-brush crash that #241 fixes; the editor dies mid-build while testing this. That is not hypothetical — it is how I hit that bug, building this PR's own test blueprint.

Closes follow-up 3 from #239 — element sensors, open since #225. Most of this work already existed on a stale branch; this rebases it onto current master, resolves what the intervening PRs changed underneath it, and re-homes its documentation.

## The feature

Seven prefabs have no threshold at all. Their setting is the mod's `Filterable` key, whose single field `SelectedTag` is an element **id string** (`"Oxygen"`) — *not* the integer hash `selected_elements` uses, which is the mistake to avoid when reading a stored blueprint.

| prefab | phase | |
|---|---|---|
| `LogicElementSensorGas`, `GasConduitElementSensor` | Gas | sensors detect |
| `LogicElementSensorLiquid`, `LiquidConduitElementSensor` | Liquid | |
| `SolidConduitElementSensor` | Solid | |
| `GasFilter`, `LiquidFilter` | Gas / Liquid | filters divert |

The phase is not a user choice — it comes from the game's own `Filterable.filterElementState` and is fixed per prefab. `FILTERABLE_BUILDINGS` maps it, `resolveSettingDescriptors` copies it onto the descriptor as `elementForceTag`, and the panel hands that to `app-cell-element-picker`'s `forceTag` (inside a `p-popover`, the same pattern as `pipe-content`).

Two states that are easy to confuse, and are deliberately distinct:

- **`NONE_TAG` (`"Void"`) rendered as "None"** — the game's own "nothing selected" default, so creating the key changes nothing until the user picks. Showing the raw `Void` was wrong because it is also a real element name and reads as a deliberate choice (that is the second commit).
- **Clear removes the whole key** — the blueprint then says nothing about the setting, which is a different claim from "set to nothing".

## What the rebase had to resolve

The branch was 16 commits behind and three of its four conflicts were substantive:

- **Its first commit was #231.** `9c04bf4d` "edit Critter Sensor settings" is the pre-review original of what shipped as #231 and was then refined by #237. Skipped in favour of master's reviewed version.
- **`threshold-sensors.ts`** — the branch's note still claimed `LogicWattageSensor` and `LogicHEPSensor` were unconfirmed carriers. #239 disproved that. Master's correction kept; only the branch's richer element-sensor wording taken.
- **`CLAUDE.md`** — the branch wrote its documentation into the monolithic root file that #230 has since split. Root file keeps master's version; the content moved to `lib/src/blueprint/CLAUDE.md`, which had a placeholder waiting for exactly this feature.

One thing the rebase got right for free, worth noting because it could easily have gone wrong: the branch refactors the stowaway-`Switch` suppression into `suppressesStowawaySwitch(prefabId)`, which delegates to `thresholdSensorSpec`. That means it automatically picked up the Wattage and Radbolt sensors #239 added *after* this branch was written, rather than hardcoding a list that would have silently gone stale.

## Verification

- Backend **1175 passing** (2 failing are the pre-existing `import-scripts` Windows path-separator assertions that pass on Linux CI); frontend **1326 passing**, 0 failures.
- `npm run build:lib` and `npm run tsc` clean.
- **Full round trip driven through a real browser.** Placed a Gas Element Sensor, selected it, opened the picker (**filtered to gases only** — Aluminum Gas, Tungsten Gas, Polluted Oxygen, Lead Gas, Nuclear Fallout, with **None** at the top), set it to Lead Gas, saved to the cloud, then loaded the blueprint fresh from its `/b/<id>` URL: the panel reads **Lead Gas** again. No console errors anywhere in that.
- **What actually got stored is the thing worth checking**, and it is right:

  ```json
  {"Key":"Filterable","Value":{"SelectedTag":"LeadGas"}}
  ```

  The element **id string**, not the integer hash `selected_elements` uses. That confusion is the specific failure this feature could plausibly have had, and it does not.
- `Set` on an unset building creates the key at the game's own `Void` default, shown as **None** with a **Clear Element** button — so creating the key changes nothing until the user picks.

- **Every phase is filtered correctly**, checked one prefab per phase:

  | building | picker lists |
  |---|---|
  | Gas Element Sensor | Aluminum Gas, Tungsten Gas, Polluted Oxygen, Lead Gas, Nuclear Fallout |
  | Liquid Element Sensor | Brackene, Liquid Methane, Molten Iron, Liquid Carbon, Molten Synthesis Gas |
  | Conveyor Rail Element Sensor | Wood, Electrum, Solid Carbon Dioxide, Cobalt, Carbon Fiber |

- **The filters behave as the different shape they are.** A selected Gas Filter's settings block is exactly `Building settings Element (Not set) Set` — no stray "On" toggle, confirming `suppressesStowawaySwitch` is the harmless no-op it is meant to be for the two prefabs that have no `Switch` component.
- **Copy carries the setting.** Set a Gas Filter to Polluted Oxygen, sampled it with **B**, placed the copy: the duplicate reads `Element: Polluted Oxygen`.

### Verified in game

@dytterud built a purpose-made blueprint in the game and the settings came through correctly. Six buildings, four configured and two controls:

| building | panel showed | stored id |
|---|---|---|
| Gas Element Sensor | Polluted Oxygen | `ContaminatedOxygen` |
| Gas Element Sensor | None | `Void` |
| Liquid Element Sensor | (Not set) | *no `buildingData` at all* |
| Conveyor Rail Element Sensor | Wood | `WoodLog` |
| Gas Filter | Oxygen | `Oxygen` |
| Liquid Filter | Water | `Water` |

Two things this settles that nothing else could:

- **The stored value is the game's internal element id, not the display name** — `ContaminatedOxygen` for "Polluted Oxygen", `WoodLog` for "Wood". Those two were chosen deliberately because their id differs from their label: if the mod resolved the string wrongly, they are where it would show. The whole `Filterable`/`SelectedTag` contract was read out of the mod source; this is the first time it has been observed applying.
- **The two controls behave.** `Void` comes through as nothing selected rather than a literal filter, which is what lets `Set` create the key as a genuine no-op; and the building with no key at all is left alone.

## Still needs testing

- Whether the panel behaves when a stored blueprint carries a `Filterable` key naming an element the current database does not know.
- Subjective: whether `cell-element-picker` inside a `p-popover` feels right in the settings panel, which is narrower than where that control is normally used. Worth a human eye rather than a screenshot.


Sinetheta:
Thanks @dytterud. This closes the element-sensor item that has been open since #225, and it is the third PR of yours to land today.

What I want to call out: the `Void` versus Clear distinction, and the second commit that stopped rendering `Void` as a literal. "Key present, nothing selected" and "no key at all" are different claims about a building, and you kept both expressible rather than collapsing them. `suppressesStowawaySwitch` delegating to `thresholdSensorSpec` is the other one; it picked up the Wattage and Radbolt sensors from #239 for free, which is what a lookup buys over a list. And the in-game round trip, with `ContaminatedOxygen` and `WoodLog` chosen precisely because their ids differ from their labels, is the check that matters here and not one CI can do.

I read the mod's `DataTransfer_Filterable` handler independently: it stores `SelectedTag.ToString()` and applies it back as a `Tag`, so the id-string contract is exactly what you described. Both suites green here (backend 1177, frontend 1326), lint clean, and pulling the seven entries from `FILTERABLE_BUILDINGS` fails the catalogue and panel specs. Merged as-is.

One correction for the record, and a follow-up sized for a single commit: the decompile has eleven `Filterable` carriers with a `filterElementState`, not seven, and four of the others are in our shipped database: `SolidFilter` (Solid) and the three `RocketInterior{Gas,Liquid,Solid}Output` buildings. Today a blueprint carrying a Solid Filter setting shows an Element row whose picker offers the default Gas and Liquid pool, and there is no Set button for it. Four lines in `FILTERABLE_BUILDINGS` and a spec that walks the database for every carrier would close it. Happy for you to take it, or I will.

Thanks again.

Issue #14, opened 2023-03-23
Spaced Out


Wizeowel:
Presuming we can add the new content... how to handle blueprints with Spaced Out buildings?
a. wild west, just add the buildings and let the user be responsible for whether it can be built in vanilla?
b. force the creator to choose a style and then restrict buildings?
c. warn about SO buildings when loading
d. fine grain the select all buildings of this type functionality so that you can e.g. select all cobalt-pipes and replace them with copper if you don't have the DLC

Ozzymud:
For A the items in question could have some indicator that they are SO when designing a blueprint.

B is not really necessary, maybe just have a toggle to clean up the selections in the build menu and if SO tools are selected auto tag the blueprint with DLC_SO

C should be a thing if there is not a filter in the browse to view vanilla, dlc, or both.

D would be awesome in the editing tool at a later date (sounds very complicated), would be awesome to have some kind of configurable replace brush... replace all cobalt pipes with copper pipes, etc.  Just drawing from the Factorio upgrade planner in game, you can pick and choose what item to replace with what item to replace it with either in blueprint mode or even after it is built.

Sinetheta:
@Wizeowel coming back to this three years late, sorry. Most of what this thread asked for has since shipped, so here is where each idea landed:

- [x] (a) Spaced Out buildings are simply in the build menu, no separate mode.
- [x] Auto-tagging: the DLCs a blueprint needs are now derived from its buildings on every save, shown as chips on the card and the details page, and nobody sets them by hand.
- [x] A browse filter for vanilla / DLC / both: Discover has "show only" and "hide" DLC facets, and a logged-in user's hide list is remembered as an account preference.
- [ ] A DLC marker on each building inside the build menu while you design: not done. Should be small, the DLC info is already on every building.
- [ ] (c) A warning when you load a blueprint that needs a DLC you don't have: not done. The site doesn't know what you own, so it would lean on that same hide preference.
- [x] (d) Replace every X with Y, the upgrade-planner idea: not done, nothing like it exists yet. The biggest of the three and probably the most useful.
- [ ] One known gap in the derivation: the game's export carries DLC info for buildings but not for elements, so a blueprint whose only DLC content is a material reads as base game. That needs a change on the export side.

Leaving this open as the to-do for the last three. Thanks for laying the options out so clearly back then, it made the eventual design easy.

PR #246, opened 2026-09-19
feat(editor): warn when an opened blueprint needs a DLC the user hides (#14 item 2)

Sinetheta:
Second of the three open to-dos from #14 (comment of 2026-09-19). Independent of #245; both branch from master.

## Not yet verified
- Manual check in the browser: log in, hide a pack on Discover, open a blueprint that needs it from `/b/<id>`, confirm the sticky warn toast. Then confirm no toast for a base-game blueprint, for a file import, and after clearing the preference. Unit specs cover the decision logic with stubs, not the real toast or the real request.
- Translation files (`src/i18n/*.xlf`) not regenerated (not done per PR in this repo).

## What shipped
- `BlueprintService.requiredDlcs`: the editor-open response's `requiredDlcs`, set only in `getBlueprint` and nulled in `reset()`, so a new blueprint or a file/share-string import can never carry the previous open's value.
- `ComponentBlueprintParentComponent.noticeHiddenDlcs()`: called at the end of `loadTemplateIntoCanvas`. For a logged-in user with a non-empty required set it reads `GET /api/users/me/dlc-preferences`, intersects, and adds a sticky `warn` toast ("Needs a DLC you hide" / "This blueprint requires Spaced Out!, which you chose to hide on Discover.") when the intersection is non-empty.
- Never fires for a logged-out visitor, an empty preference, an empty or absent required set. Never writes the preference. A failed read is swallowed. If the user opened a different blueprint before the preference came back, the notice is dropped rather than shown against the wrong blueprint.

## Design decisions
- Sticky warn toast, matching the existing "Unrecognized buildings" notice in the same method, rather than a banner or modal. It is a caveat, not a block.
- One request per server open, no caching. The preference can change on Discover between opens, and the request is small.
- The signal is the server's `requiredDlcs`, per the issue, not a recompute from the loaded buildings. That is why imports get no notice: they have no server-derived set. A follow-up could derive it client-side with lib `deriveRequiredDlcs` if imports should warn too.
- Spec: added a separate, non-skipped `describe` for the notice with every collaborator stubbed and no `detectChanges`; the existing skipped `describe` is untouched. The component-level `MessageService` provider is overridden via `TestBed.overrideComponent`.

## Verified
```
cd frontend && npm test -- --include='**/component-blueprint-parent.component.spec.ts'
 ✓ 7 passed | 2 skipped (9)        [the 2 skipped are the pre-existing skipped suite]
npm test -- --include='**/blueprint-service.spec.ts'   → 102 passed
npx tsc --noEmit -p tsconfig.app.json                   → clean
ng lint                                                 → all files pass
```


Sinetheta:
@coderabbitai review

Sinetheta:
@coderabbitai full review

PR #247, opened 2026-09-19
docs(spec): design note for replace-every-X-with-Y (#14 item 3)

Sinetheta:
Third of the three open to-dos from #14 (comment of 2026-09-19). **Design note only, no code.** Waiting on decisions before implementation.

## Not verified
- Nothing runs here; it is one Markdown file. The `spec/` directory did not exist in this checkout (only gitignored `/specs/`), so this creates it. If the notes referenced from CLAUDE.md as `spec/...` live somewhere else, say where and I will move it.

## What the note covers (`spec/replace-with-plan.md`)
- **What exists**: the per-group element picker already does an unconditional material swap per `SameItemCollection`, one undo step per group; the Element Report's "select every X" already produces the starting selection. The feature is a cross-group, conditional, single-undo-step version of that plus a control.
- **Scope**: recommends v1 = material swap only (slot-wise X to Y, skip buildings that cannot take Y, never force). Building swap as v2 restricted to allow-listed families (liquid pipe, gas pipe, wire, 1x1 tile), ids checked against `database-2024.json`.
- **Control**: a Replace strip in the Select Tool card acting on the current selection; rejects putting it on Element Report lines (would silently mean "everywhere").
- **Undo**: one step via `pauseChangeEvents` / `resumeChangeEvents(true)`; selection re-collected after apply.
- **buildingData**: untouched by a material swap, by construction. For a building swap, carry over only keys the settings catalogue says apply to the target; drop and count the rest; never synthesize defaults.

## Decisions needed
1. v1 = material swap only? (recommended yes)
2. Control in the Select Tool card, acting on the selection? (recommended yes)
3. Building swap later: allow-listed families vs generic compatibility check? (recommended families)


Sinetheta:
Closing: design accepted in chat (material swap only, control in the Select Tool card, one undo step). Implementation prompt handed off outside the repo; nothing to-do-shaped gets checked in.

